PR #138, opened 2014-10-27
Arietta

nzmichaelh:
The Arietta is a AT91SAM9G25 development board that runs Linux.  This patch adds support for the board, digital I/O, PWM, and the I2C bus.

I'd like to turn this into a generic Linux driver next and perhaps merge the BeagleBone, Edison, and Arietta drivers into one.

### DIFF
--- a/examples/arietta_blink.go
+++ b/examples/arietta_blink.go
@@ -1,0 +1,32 @@
+package main
+
+import (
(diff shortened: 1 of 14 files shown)


deadprogram:
This PR is just too far out of date, despite being interesting. Closing.

PR #179, opened 2015-03-17
Add a Stream type that handle the request to /events/:event and write tests for event stream

rafmagana:
* Test coverage for event stream
* Fixes json error when event doesn't exists

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -224,15 +224,12 @@ func (a *API) robotDevice(res http.ResponseWriter, req *http.Request) {
 	}
 }
 
-// robotDeviceEvent returns device event route handler.
-// Creates an event stream connection
-// and queries event data to be written when received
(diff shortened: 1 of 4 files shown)


rafmagana:
@zankich  The time the test suite takes to run was increased more than 2 seconds, don't know if that's a problem for Travis because it doesn't say what the problem was, I didn't find another way to test the events stream, I'll check this out tomorrow morning.

rafmagana:
@zankich a test is failing in Go 1.2, I only tested locally with Go 1.3, so I'm going to install go 1.2 and see what's the problem.

rafmagana:
@zankich I added another way to test the `/events/:event` endpoint, I realized that in the first way I did it I was only testing if the value was being sent to the channel not the actual response, in my last commit I test the actual response (and it works with Go 1.2), please, check it out and let me know.

zankich:
:shaved_ice:

PR #184, opened 2015-04-16
platforms/leap: use golang.org/x/net/websocket

sbinet:
- `code.google.com/p/...` will disappear.
- use the new home for the `go.xyz` sub repos


zankich:
Thanks!

PR #186, opened 2015-05-08
sysfs seek

stengaard:
Use Seek in package sysfs instead of doing Open/Close dance on every Read and Write of digitalPins.

### DIFF
--- a/platforms/intel-iot/edison/edison_adaptor_test.go
+++ b/platforms/intel-iot/edison/edison_adaptor_test.go
@@ -55,6 +55,7 @@ func initTestEdisonAdaptor() (*EdisonAdaptor, *sysfs.MockFilesystem) {
 		"/sys/class/gpio/gpio221/direction",
 		"/sys/class/gpio/gpio243/value",
 		"/sys/class/gpio/gpio243/direction",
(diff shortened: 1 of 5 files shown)


deadprogram:
Hi, @stengaard sorry I have not followed up on this PR. These days the work in progress branch is the `dev` branch, so this PR should be submitted against that branch. Can you please look at doing that, sinec it looks like a great thing to have!

Any opinion from @trevrosen or @mattetti on this code?

trevrosen:
@deadprogram I'm a big fan of this change. I'm also wondering if we shouldn't make the `dev` branch requirement higher up/bolded/etc to make it more prominent in `CONTRIBUTING.md`, because you have to read pretty far down to see that and we've had several high-value PRs that were originally opened against `master`

@stengaard, I understand that it's been a long time since you've made this PR and you may have moved on to other things. If you'd be so kind as to re-open against `dev`, I have a feeling that @deadprogram will merge it quickly. If not, we thank you for the contribution anyway and will probably move forward remaking the PR ourselves. In any case, thanks for your interest in and contributions to Gobot!

stengaard:
Ok, I'll go ahead and re-open against `dev`.

PR #192, opened 2015-07-02
Add support for 'hand' and 'gesture' Leap Motion events

stewart:
Adds :hand: + :wave: support.

Unsure if this is the best way to add them, let me know if there's a possible refactoring I'm missing.

### DIFF
--- a/platforms/leap/leap_motion_driver.go
+++ b/platforms/leap/leap_motion_driver.go
@@ -24,6 +24,8 @@ var receive = func(ws io.ReadWriteCloser, msg *[]byte) {
 //
 // Adds the following events:
 //		"message" - Gets triggered when receiving a message from leap motion
+//		"hand" - Gets triggered per-message when leap motion detects a hand
+//		"gesture" - Gets triggered per-message when leap motion detects a hand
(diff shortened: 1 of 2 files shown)


deadprogram:
That looks about like what I expected. Example or 2 please?

deadprogram:
Also, do we have this documented in the godoc?

platforms/leap/leap_motion_driver.go
```diff
@@ -60,9 +62,19 @@ func (l *LeapMotionDriver) Start() (errs []error) {
...
 		var msg []byte
+		var frame Frame
 		for {
 			receive(l.adaptor().ws, &msg)
-			gobot.Publish(l.Event("message"), l.ParseFrame(msg))
+			frame = l.ParseFrame(msg)
+			gobot.Publish(l.Event("message"), frame)
+
+			for _, hand := range frame.Hands {
+				gobot.Publish(l.Event("hand"), hand)
```
zankich: Do we want an event for each hand or one event with all the hands?

deadprogram:
This PR is no longer valid, being that it is against master instead of dev. Closing.

deadprogram:
I merged directly into `dev` branch, as the tests appeared to be passing.

PR #194, opened 2015-07-08
Add missing package install step to Leap Motion README

deadprogram:
Added missing step to docs.


deadprogram:
Updated to include all current platforms.

stewart:
:books:

PR #196, opened 2015-07-09
Add basic keyboard driver using `stty`

stewart:
Won't work on Windows unfortunately, but should work on most Unix systems.

### DIFF
--- a/examples/keyboard.go
+++ b/examples/keyboard.go
@@ -1,0 +1,36 @@
+package main
+
+import (
+	"fmt"
+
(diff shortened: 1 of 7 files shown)


deadprogram:
Works on OSX?

stewart:
That's what I tested it on, would like for Linux users to test it as well though

deadprogram:
Tried on Linux, it worked, however, after terminating the example, the keys I press are no longer displayed in my terminal.

deadprogram:
Looks good now, thanks @stewart

PR #202, opened 2015-07-10
jhd1313m1_driver: fix '\n' processing

toxeus:
Thanks to the guys at the other table :+1:

### DIFF
--- a/platforms/i2c/jhd1313m1_driver.go
+++ b/platforms/i2c/jhd1313m1_driver.go
@@ -34,6 +34,8 @@ const (
 	LCD_MOVERIGHT           = 0x04
 	LCD_MOVELEFT            = 0x00
 	LCD_2LINE               = 0x08
+
+	LCD_2NDLINEOFFSET = 0x40
 )
 
 var _ gobot.Driver = (*JHD1313M1Driver)(nil)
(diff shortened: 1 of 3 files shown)


platforms/i2c/jhd1313m1_driver.go
```diff
@@ -145,6 +147,12 @@ func (h *JHD1313M1Driver) Home() (err error) {
 
 func (h *JHD1313M1Driver) Write(message string) (err error) {
 	for _, val := range message {
+		if val == '\n' {
+			if err = h.SetPosition(16); err != nil {
+				return
+			}
+			continue
+		}
 		if err = h.connection.I2cWrite(h.lcdAddress, []byte{0x40, byte(val)}); err != nil {
```
mattetti: nitpick, shoudn't we use `LCD_2NDLINEOFFSET` instead of `0x40` in the `I2cWrite` call?

mattetti:
+1 for this PR, ideally we would also refactor a few places where we use `0x40` and switch to using the newly defined constant.

platforms/i2c/jhd1313m1_driver.go
toxeus: No, I don't think so. In this context it should be `LCD_SETCGRAMADDR`.

toxeus:
> ideally we would also refactor a few places where we use `0x40`

IMHO we should use constants instead of plain values everywhere. Let's clean this up in a dedicated PR :wink:

zankich:
Closing this PR due to https://github.com/hybridgroup/gobot/pull/209

PR #209, opened 2015-07-12
Fixes LCD setup + wait times to avoid bugs + refactoring

mattetti:
The refactoring focuses on 2 aspects:
* go lint conventions with documentation
* avoid naked returns and don't instantiate returned values in the signature unless really necessary

I probably should have split this PR in two.

### DIFF
--- a/platforms/i2c/jhd1313m1_driver.go
+++ b/platforms/i2c/jhd1313m1_driver.go
@@ -35,6 +35,7 @@ const (
 	LCD_MOVELEFT            = 0x00
 	LCD_2LINE               = 0x08
 	LCD_CMD                 = 0x80
+	LCD_DATA                = 0x40
 )
 
 var _ gobot.Driver = (*JHD1313M1Driver)(nil)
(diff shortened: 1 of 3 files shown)


mattetti:
also merged and rebased #202 onto this PR

platforms/i2c/jhd1313m1_driver.go
```diff
@@ -34,10 +34,18 @@ const (
 	LCD_MOVERIGHT           = 0x04
 	LCD_MOVELEFT            = 0x00
 	LCD_2LINE               = 0x08
+	LCD_CMD                 = 0x80
+	LCD_DATA                = 0x40
```
toxeus: Please use `LCD_SETDDRAMADDR` instead of `LCD_CMD` and `LCD_SETCGRAMADDR` instead of `LCD_DATA`.

mattetti: I chose to use the constants used in the C++ code so it's easier to compare code changes

zankich:
Thanks for the pull request, testing now....

zankich:
@mattetti I was able to fix the `Scroll` method locally, so I will push up my fix once I merge. I am also going to make a change to the `Write` function so  that it will  accept the `io.Writer` interface and create a new `WriteString` function  which will accept a string.

Thanks @mattetti and @toxeus!

mattetti:
I'm looking forward to seeing how you fixed the scrolling issue

platforms/i2c/jhd1313m1_driver.go
toxeus: I see. Rather then copying the constant names from the C++ code they should be chosen such that they match the [datasheets](http://www.seeedstudio.com/wiki/images/0/03/JHD1214Y_YG_1.0.pdf). The whole point of Gobot is to enable developers to write code for embedded devices in Go instead of C++.

Now, we have 3 different styles in the code:
* Gobot constants that match the spec
* constants taken from the C++ code
* plain values

But as long as the maintainers are fine with that then I'm too :wink:

deadprogram: I agree with @toxeus we should use the labels from datasheet.

PR #214, opened 2015-07-20
Odroid c1

dhart-alldigital:
Added support for odroid-c1.  Got the normal gpio and pwm working.  I'm not a go programmer, this is my first time at using the language, so feel free to provide any feedback,


zankich:
@dhart-alldigital thanks for the pull request! I've been thinking about this PR for the past few days and how I want to approach it's inclusion. Since I don't have an Odroid and I'm not sure if/when I'll get one, it will be hard to support this package. The options which I've come up with is either you can create a repo for this Odroid code and I'll link it on the Gobot readme/website or I'll create a `Gobot-contrib` repo and give you, and anyone else who wants it, commit access over the repo.

What are your thoughts?

dhart-alldigital:
I like the idea of a dedicated `Gobot-contrib` project rather than a standalone repo that would need to be maintained just for the odroid.  We could keep the contrib project streamlined to just providing extensions to the `platforms` package (plus `examples`).  A clear cut separation of concerns.

deadprogram:
Cool looking platform, please update or submit a new PR if you would like to add support to Gobot 1.0. Thanks.

PR #215, opened 2015-07-20
Save ~1000 goroutines

joushou:
Previous code spawns ~1000 go-routines due to NewEvent in my test. Removing a
goroutine and making the loop of callbacks happen directly in Write with
a lock reduces this to ~12 in my test.

While I agree that explicit locks are not nearly as nice looking as implicitly locking channels, I think the gain in simpler stack traces (~1000 x a few lines gives a LONG trace!) is enough of a reason by itself to include this change. I see no performance penalty of interest (the amount of locks have not been reduced - every channel is a locked queue, and the time to loop over a slice that often has a count of 1 is negligible), but there might be a small performance gain overall, due to lower count of goroutines and active locks that the scheduler has to keep track of. That is, however, just a bold and unscientific claim.

### DIFF
--- a/event.go
+++ b/event.go
@@ -1,5 +1,7 @@
 package gobot
 
+import "sync"
+
 type callback struct {
 	f    func(interface{})
 	once bool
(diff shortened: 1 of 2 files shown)


explicite:
:thumbsup:

trevrosen:
:+1: 

This would solve issues I've had in making large-scale refactors to the `api` internal package -- the 80+ goroutines spawned in a normal run of `api_test.go` make any panic in there annoying.

Now that we have `go test trace` in 1.5, I've traced tests of `api_test.go` and that's it. I agree with @joushou that this might not make for any kind of huge perf gain (tho it will likely be non-zero), but the big win is easier debugging of stuff that implicates the event system.

deadprogram:
@joushou can you please resubmit this PR against the `dev` branch?

joushou:
Done

PR #220, opened 2015-08-21
Update pin assignments to be correct for my beaglebone.

aliafshar:
e.g. http://robotic-controls.com/node/68


deadprogram:
Thanks, @aliafshar

PR #225, opened 2015-08-28
Wrong analog input mapping for BeagleBone

pruggia:
According to https://github.com/CircuitCo/BeagleBone-Black/raw/master/BBB_SRM.pdf Page 105 these are the right mappings:
P9_36 -> AIN5
P9_35 -> AIN6
Thanks!


deadprogram:
Thank you, @pruggia

PR #228, opened 2015-09-09
Added Firmata Reset function to firmata adaptor.

joek:
Hi,
I need the Reset function out of the Firmata Protocoll/Client available through the firmata adaptor. Since it is already implemented on Client side, it might be a good idea to add this on plattform level too.

### DIFF
--- a/platforms/firmata/firmata_adaptor.go
+++ b/platforms/firmata/firmata_adaptor.go
@@ -1,6 +1,7 @@
 package firmata
 
 import (
+	"fmt"
 	"io"
 	"strconv"
 	"time"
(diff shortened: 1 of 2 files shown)


deadprogram:
Hi, @joek can you please resubmit this PR against the `dev` branch? Thank you!

deadprogram:
Hi, @joek this looks good, can you please resubmit against the `dev` branch? 

We added some guidelines for contributors that explains why: https://github.com/hybridgroup/gobot/blob/dev/CONTRIBUTING.md#general-guidelines

Thank you!

joek:
@deadprogram sorry for the delay, but I am on parentel leave and don't got frequent access to the internet. The Problem is, the change is not working... The function is only sending an reset to the board. It is not resetting all the values in the go lib. So e.g. a servo pin is still in servo mode inside the go routine but reset on arduino level. 

I will close the ticket, because I don't get time to fix it at the moment.

PR #235, opened 2015-09-30
Save ~1000 goroutines

joushou:
Previous code spawns ~1000 go-routines due to NewEvent in my test. Removing a
goroutine and making the loop of callbacks happen directly in Write with
a lock reduces this to ~12 in my test.

While I agree that explicit locks are not nearly as nice looking as implicitly locking channels, I think the gain in simpler stack traces (~1000 x a few lines gives a LONG trace!) is enough of a reason by itself to include this change. I see no performance penalty of interest (the amount of locks have not been reduced - every channel is a locked queue, and the time to loop over a slice that often has a count of 1 is negligible), but there might be a small performance gain overall, due to lower count of goroutines and active locks that the scheduler has to keep track of. That is, however, just a bold and unscientific claim.

(This is #215 reopened on dev)

### DIFF
--- a/event.go
+++ b/event.go
@@ -1,5 +1,7 @@
 package gobot
 
+import "sync"
+
 type callback struct {
 	f    func(interface{})
 	once bool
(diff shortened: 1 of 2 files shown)


deadprogram:
Thanks @joushou can you please rebase against `dev` just one more time? The build there is now "fixed". Thanks!

deadprogram:
Also, @mattetti and anyone else, any opinion about this PR?

mattetti:
LGTM, besides saving a lot of goroutines (which mainly affect stack traces as mentioned by @joushou ), I do find the new code more readable. It's a :+1: for me, thanks Kenny!

deadprogram:
Looking good, merging.

PR #237, opened 2015-10-08
Update blink example due to API change

mattetti:
the existing example doesn't compile since the API changed a while back, this PR fixes the issue.

### DIFF
--- a/platforms/firmata/client/examples/blink.go
+++ b/platforms/firmata/client/examples/blink.go
@@ -14,10 +14,10 @@ func main() {
 		panic(err)
 	}
 
-	board := client.New(sp)
+	board := client.New()
 
 	fmt.Println("connecting.....")
-	err = board.Connect()
+	err = board.Connect(sp)
 	defer board.Disconnect()
 
 	if err != nil {


deadprogram:
This was overlooked, in part due to most of the examples having been put into the `/examples` directory. Having examples in both locations is a bit confusing, I think. Any thoughts on that?

mattetti:
I find it nice to be able to get examples close to the code I care about. I wanted to use the low level version of the library for a talk I'm giving. I wouldn't have thought about going up to another package to find an example just for this package. But I might be misunderstanding the situation

deadprogram:
OK how about we have platform specific examples under platforms, and multi platform can just be under examples, and put a README in both places mentioning more examples can be found in the other location.

mattetti:
Just to be clear, there are 2 blink examples:

* https://github.com/hybridgroup/gobot/blob/dev/examples/firmata_blink.go
and 
* https://github.com/hybridgroup/gobot/blob/dev/platforms/firmata/client/examples/blink.go

The first one is higher level and uses and adaptor and the `gpio` package when the second one uses the board directly.
I think we should keep both examples since they are very valuable.

I think you're suggesting to have an examples directory in the platforms folder (at the same level as all the platform folders). Inside this directory we would move the second example shown above (inside a nested folder called firmata) and also add a README to mention that at the root level, there are more high level examples.
If my understanding of your suggestion is right, I agree that it's a good idea.

trevrosen:
:+1:

deadprogram:
That was not exactly what I was suggesting @mattetti but I like your idea better.

deadprogram:
I will merge this PR, then we can deal with moving around the lower-level example in a separate commit/series of commits as needs be. Thanks!

PR #238, opened 2015-10-11
let devs set and use custom LCD characters

mattetti:
I noticed that the grove LCD screen only has an ASCII/Japanese font table. I need to be able to display accented characters looking at the data sheet the screen supports up to 8 custom CGRAM characters. I added an API to set those CGRAM characters as well as a map of common latin 1 characters + emoticons


mattetti:
I'm pretty happy with the end result, it took me a little while to figure out how to manage custom characters but I think the result is pretty straight forward and useful. The only downside is that developers have to load the custom chars they want and they need to edit their display text on the fly to swap custom characters for their positions. As @deadprogram knows, not having accented characters such as ñ is Spanish can be extremely problematic ;)

deadprogram:
:european_post_office: muy bien, gracias!

PR #243, opened 2015-11-07
Fixed event race condition

cathalgarvey:
Relates to the example given for the Parrot AR2 Drone.

If this is anything like the Bebop, then taking off immediately fires the "flying" event, but these events aren't buffered so attaching an event handler afterwards achieves nothing. When I ran the same code for the Bebop (as suggested on the site) it never landed, prompting an emergency connect-with-phone-hammer-buttons scramble. Swapping the event registration with the takeoff method call fixed the problem immediately.


deadprogram:
Thanks, @cathalgarvey

PR #251, opened 2016-01-02
encoder driver implementation

mattetti:
the data coming back from my arduino doesn't seem to be quite correct and I get forward and backward data confused in some cases (the pin value doesn't seem consistent, it might be the symptom of a deeper bug)

### DIFF
--- a/platforms/gpio/encoder_driver.go
+++ b/platforms/gpio/encoder_driver.go
@@ -1,0 +1,110 @@
+package gpio
+
+import (
+	"strconv"
+	"time"
+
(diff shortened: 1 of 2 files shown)


mattetti:
More info about encoders and why I think my reading is off: http://playground.arduino.cc/Main/RotaryEncoders

mattetti:
To do this properly, we need to be able to use interrupts which we can't with the current version of firmata used by gobot. Ideally we would be moving to confgurable firmata and load an encoder wrapper such as: https://github.com/firmata/FirmataEncoder

PR #253, opened 2016-01-09
Correct the release command sent to pi-blaster.

ekyoung:
The release command sent to pi-blaster from Finalize in the RaspiAdaptor uses the wrong syntax. As a result, pi-blaster prints "Bad input: ..." to STDERR and continues to apply the last PWM value written to the pin.

Code demonstrating the issue: https://github.com/ekyoung/gobot-pi-blaster-release-issue


deadprogram:
Great work, @ekyoung thank you very much! Merging.

PR #260, opened 2016-02-07
Get I2C functionality before doing SMBus block I/O

hrishikesh195:
In the sysfs i2cDevice implementation, use an ioctl to get the adapter
functionality mask. Prefer SMBus block I/O but if it's not available,
perform read/write calls directly on the file descriptor.

Improve Wiichuck error handling. Add a 1 ms delay between I/O operations
to the Wiichuck; this dramatically improves reliability.

Signed-off-by: Hrishikesh Tapaswi <hrishikesh195@yahoo.com>

### DIFF
--- a/examples/chip_wiichuck.go
+++ b/examples/chip_wiichuck.go
@@ -26,6 +26,9 @@ func main() {
 		gobot.On(wiichuck.Event("z"), func(data interface{}) {
 			fmt.Println("z")
 		})
+		gobot.On(wiichuck.Event("error"), func(data interface{}) {
+			fmt.Println("Wiichuck error:", data)
+		})
 	}
 
 	robot := gobot.NewRobot("chuck",
--- a/examples/firmata_wiichuck.go
+++ b/examples/firmata_wiichuck.go
@@ -26,6 +26,9 @@ func main() {
 		gobot.On(wiichuck.Event("z"), func(data interface{}) {
 			fmt.Println("z")
 		})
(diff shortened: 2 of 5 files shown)


deadprogram:
Thanks @hrishikesh195 for working on this. Could you please look into the failing test here: https://travis-ci.org/hybridgroup/gobot/jobs/107532507#L665

hrishikesh195:
I think the assertion in the test is incorrect. The assertion should expect 3 bytes, not 4 since only 3 were previously written -

```go
	buf := []byte{0x01, 0x02, 0x03}

	n, err := i.Write(buf)
```

So even though the buffer used for reading has size 4, the read should return the number of bytes actually read.

The test did not fail earlier because the function `func (d *i2cDevice) Read()` would stash away the length of the passed-in buffer in the first byte of the buffer given to ioctl. Since the mock syscall is a no-op, that value would remain unchanged and would end up as the return value.

I'll change the test assertion and update the PR.

hrishikesh195:
Do you think you could also test this change with your Grove LCD just as a regression test? Thanks.

deadprogram:
Hi, @hrishikesh195 my LCD works perfectly. I'm looking for one of my i2c devices that is r/w just to be sure... I'll get back to you shortly.

deadprogram:
Tried with my MMA7660 and works perfectly. Thank you so much for your fine work here, @hrishikesh195 this is great. Merging.

hrishikesh195:
My pleasure!

PR #261, opened 2016-02-08
Name C.H.I.P. pins according to printed names

hrishikesh195:
It's more convenient to refer to a pin by the name that's printed right
on the header itself instead of having to count the pin number.

Signed-off-by: Hrishikesh Tapaswi <hrishikesh195@yahoo.com>

### DIFF
--- a/examples/chip_blink.go
+++ b/examples/chip_blink.go
@@ -12,7 +12,7 @@ func main() {
 	gbot := gobot.NewGobot()
 
 	chipAdaptor := chip.NewChipAdaptor("chip")
-	led := gpio.NewLedDriver(chipAdaptor, "led", "U14_13")
+	led := gpio.NewLedDriver(chipAdaptor, "led", "XIO-P0")
 
 	work := func() {
 		gobot.Every(1*time.Second, func() {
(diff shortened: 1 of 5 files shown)


hrishikesh195:
Sorry I should have picked the right naming scheme for the pins in the initial commit itself. I had based it off of the Beaglebone Black naming scheme but I realized that the C.H.I.P. has pin names printed right on the header itself and it would be better to use those names.

hrishikesh195:
I'm not sure why this test failed. The changes seem unrelated to the makey button. I'm wondering if this is a timing issue. In the test, it seems like one goroutine is expected to receive something from a recently spawned goroutine within 15 ms. I'm wondering if the spawned goroutine didn't get a chance to run at all, thereby causing the timeout.

Is there a way to trigger a re-run of this test?

deadprogram:
Thanks for this update, @hrishikesh195 that is how we have tried to do the pin mapping on other boards. The "PX_Y" format is specific to the Beaglebone Black, hence why we used that system.

I restarted the Travis build, looks like we need a small tweak to the tests to avoid that timing problem.

deadprogram:
I am going to merge this, since the current Travis build failure is not related. Great work, @hrishikesh195 thank you!

PR #264, opened 2016-02-18
Use Seek to speed up read/write in sysfs

stengaard:
This maintains `direction` and `value` `File`s for each DigitalPin
implementation. Instead of Open/Read/Close we now only do Seek/Read,
this speeds up Read/Write operations a bit.

A silly benchmark on the mock FS gives:

benchmark                  old ns/op     new ns/op     delta
BenchmarkDigitalRead-8     647           7.36          -98.86%

benchmark                  old allocs     new allocs     delta
BenchmarkDigitalRead-8     5              0              -100.00%

benchmark                  old bytes     new bytes     delta
BenchmarkDigitalRead-8     96            0             -100.00%

### DIFF
--- a/platforms/intel-iot/edison/edison_adaptor_test.go
+++ b/platforms/intel-iot/edison/edison_adaptor_test.go
@@ -65,6 +65,7 @@ func initTestEdisonAdaptor() (*EdisonAdaptor, *sysfs.MockFilesystem) {
 		"/sys/class/gpio/gpio221/direction",
 		"/sys/class/gpio/gpio243/value",
 		"/sys/class/gpio/gpio243/direction",
+		"/sys/class/gpio/gpio229/value",
 		"/sys/class/gpio/gpio229/direction",
 		"/sys/class/gpio/gpio253/value",
 		"/sys/class/gpio/gpio253/direction",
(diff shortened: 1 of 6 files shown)


stengaard:
@trevrosen @deadprogram here's the rewritten version. Feedback is of course more than welcome!

trevrosen:
@stengaard can you repro Travis' test failure for the Edison adapter locally?

stengaard:
Yeah, I see the same problem popped up in the original PR.

trevrosen:
Yeah that would be a regression... :crying_cat_face:  If you can put together a fix, pretty sure @deadprogram would be willing to merge tho!

stengaard:
@trevrosen - should be fixed by f839367

trevrosen:
:shipit:

deadprogram:
This is really cool, @stengaard I'm plugging in a few devices to give it some "real-world" testing right now...

deadprogram:
OK, so far I am extremely impressed by the performance boot. I do not have the formal benchmarks that you did @stengaard but anecdotally it is notable.

I am merging this with great pleasure, thanks @trevrosen for you help in landing it, as well.

trevrosen:
@deadprogram happy to help!

PR #265, opened 2016-02-20
Make dev branch target more explicit in CONTRIBUTING

trevrosen:
We've had some high-value PRs lag a bit because they weren't originally opened against `dev` as CONTRIBUTING specifies. Let's make that more explicit.

Making this PR against `master` so that it will be more easily seen. Might oughta be back ported to `dev` too, but I leave that to @deadprogram.


deadprogram:
:book: thanks @trevrosen

trevrosen:
Shizzam! That was fast! :100:

PR #267, opened 2016-02-21
Fix for analog lag

kittuov:
### Issue:
Lag occurs when events are produced faster than poll interval. 
### Reproduce the issue:
run the example code with analog joystick and move the analog sticks analog sticks. Two sticks at a time produce even more prominent lag.
### Fix:
finish all the stacked events at once and sleep through the interval


kittuov:
the build failed about some button driver no related to above changes i suppose

deadprogram:
Thank @kittuov looks good, and works for me. Merging.

The test fail is indeed a spurious error with one of the GPIO tests that I need to look into.

PR #268, opened 2016-02-22
code depending on "testing" package must be in _test.go files

gmarik:
Otherwise any code importing `gobot` package will have
testing flags as well:

### Problem
Unexpected flags in my app:
```
     -test.bench string
       	regular expression to select benchmarks to run
     -test.benchmem
       	print memory allocations for benchmarks
```

### Reproduction

Here's a minimum code to reproduce the problem:

```Go
package main
// main.go

import (
	"flag"
	"github.com/hybridgroup/gobot"
)

var _ = gobot.NewJSONDevice

func main() {
	flag.Parse()
	flag.PrintDefaults()
}
```

#### Output
```console
flag_test$ go run main.go
  -test.bench string
    	regular expression to select benchmarks to run
  -test.benchmem
    	print memory allocations for benchmarks
  -test.benchtime duration
    	approximate run time for each benchmark (default 1s)

# .....
````
## Fix
Extract test helpers to `gobot/gobottest` package.

### DIFF
--- a/api/api_test.go
+++ b/api/api_test.go
@@ -12,6 +12,7 @@ import (
 	"time"
 
 	"github.com/hybridgroup/gobot"
+	"github.com/hybridgroup/gobot/gobottest"
 )
 
 func initTestAPI() *API {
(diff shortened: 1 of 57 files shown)

### CODE REVIEW

**Comment by deadprogram** [comment]
Hi, @gmarik I like the concept, but as you can tell https://travis-ci.org/hybridgroup/gobot/jobs/110869133 that are many tests that depend on the current setup.

If you want to run all the tests on your machine, use the command `make test`

Perhaps a `gobot/gobottest` subdirectory would be a nice way to implement this refactoring. 

Many tests would need to be rewritten from this change, however, sounds like that is the case no matter what, at this point.

What do you think?

**Comment by gmarik** [comment]
@deadprogram yeah, i've noticed. Took me a bit to get tests failing.

Hacked together a quick-fix, but if you think we should do it the right way with `gobottest` package i'll rewrite it. Let me know.

**Comment by deadprogram** [comment]
A separate test package would be so much nicer, if you please, and thank you!

**Comment by gmarik** [comment]
@deadprogram done!

Also please double check your latest merge is successfully fixed (as it introduced conflict on my side)

**Comment by gmarik** [comment]
Any ideas why Is this failing?
```
--- FAIL: TestPublish (0.00s)
	gobottest.go:12: utils_test.go:51: 4 - "int", should equal,  1 - "int"
```

deadprogram:
Not sure about the first fail, but the second is due to the mqtt package requires versions of Go >= 1.3, which is why 1.2 is in the Travis allows failure section: https://github.com/hybridgroup/gobot/blob/dev/.travis.yml#L13

gmarik:
fixed with #269

gmarik:
@deadprogram please review! 
Merged fix to this PR from #269

deadprogram:
This looks like the right idea to me. Any feedback @trevrosen or @mattetti or anyone else?

trevrosen:
:+1: 

Really happy to see some of this test stuff getting love. This will help me with some long-time goals that I'm woefully behind on. :smiley_cat: 

:shipit:

deadprogram:
:mag:

gmarik:
Thanks guys!

deadprogram:
Thanks to you, @gmarik

PR #269, opened 2016-02-23
Fix broken tests

gmarik:
*Problem*

Since `errFunc` is a variable it's crucial to ensure its value is
properly set.

```go
var errFunc = func(t *testing.T, message string) {
    t.Errorf(message)
}
```

but `TestAssert` and `TestRefute` happen to override it

```go
errFunc = func(t *testing.T, message string) {
    err = message
}
```

without restoring the previous value, causing all the following tests to never
fail as, instead of calling `t.Errorf`, it just assigns `err` value

*Solution*

This commit refactors `Assert` and `Refute` to not use global `errFunc`
variable, but instead introduce a `TestFailer` interface to achieve
same results

### DIFF
--- a/utils.go
+++ b/utils.go
@@ -10,7 +10,6 @@ import (
 	"reflect"
 	"runtime"
 	"strings"
-	"testing"
 	"time"
 )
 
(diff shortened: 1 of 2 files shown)


gmarik:
this also takes care of #268

gmarik:
I did change signature of `Refute`  | `Assert`. If it's not desired i'll have `nopFailer` to handle the message

deadprogram:
I liked the `gobottest` package, and that seems more the correct pattern in Golang, no?

gmarik:
Ok, closing this one, in favour of #268. This attempt to figure out why tests were failing.

PR #273, opened 2016-03-28
Add MQTT authentication support

nobodyzzz:


deadprogram:
Thanks @nobodyzzz looks good. We will probably add a few function like this to allow for certificate-based auth as well.

PR #274, opened 2016-04-01
Added example of how to use temp36 temperature sensor with firmata

jeromefroe:
Hi! I just started using Gobot (it's great!) and I wrote a program to read the temperature from a temp36 sensor with an arduino that I thought might serve as a good example. It's essentially just a translation of [this example from Adafruit](https://learn.adafruit.com/tmp36-temperature-sensor/using-a-temp-sensor) into Go using Gobot. When I was looking through the examples though, I didn't see any that used the `AnalogRead` function, so I thought it might be helpful to include one. Thanks!

### DIFF
--- a/examples/firmata_temp36.go
+++ b/examples/firmata_temp36.go
@@ -1,0 +1,43 @@
+package main
+
+import (
+	"time"
+
+	"fmt"
+
+	"github.com/hybridgroup/gobot"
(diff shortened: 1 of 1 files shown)


deadprogram:
Hi @jeromefroe that looks great, thank you very much! Merging...

PR #278, opened 2016-05-01
BME280 driver

dsockwell:
I wrote a driver for the Bosch BME280 humidity/temperature/pressure sensor over I2C. 

Also in this branch is a one-line fix for the Beaglebone PWM driver causing my power supply to audibly sing at 2KHz. I switched the frequency to 20KHz. I have no basis on which to suspect this is a good thing, but it's quieter and hasn't burned my particular MOSFET setup yet.

Edit: Off by a factor of 10.

### DIFF
--- a/platforms/i2c/bme280_driver.go
+++ b/platforms/i2c/bme280_driver.go
@@ -1,0 +1,333 @@
+package i2c
+
+import (
+	"github.com/hybridgroup/gobot"
+
+	"bytes"
+	"encoding/binary"
+	"time"
+)
+
+const bme280Address = 0x76
+
+var _ gobot.Driver = (*BME280Driver)(nil)
+
+const BME280_REGISTER_PRESSURE_MSB = 0xF7
+const BME280_REGISTER_PRESSURE_LSB = 0xF8
+const BME280_REGISTER_PRESSURE_XLSB = 0xF9
+
+const BME280_REGISTER_TEMP_MSB = 0xFA
+const BME280_REGISTER_TEMP_LSB = 0xFB
+const BME280_REGISTER_TEMP_XLSB = 0xFC
+
+const BME280_REGISTER_HUM_MSB = 0xFD
+const BME280_REGISTER_HUM_LSB = 0xFE
+
+const BME280_PRESSURE_TEMPERATURE_CALIB_DATA_LENGTH = 26
+const BME280_HUMIDITY_CALIB_DATA_LENGTH = 7
+
+const BME280_DATA_FRAME_LENGTH = 8
+
+const BME280_REGISTER_DIG_T1_LSB = 0x88
+const BME280_REGISTER_DIG_T1_MSB = 0x89
+
+const BME280_REGISTER_DIG_T2_LSB = 0x8A
+const BME280_REGISTER_DIG_T2_MSB = 0x8B
+
+const BME280_REGISTER_DIG_T3_LSB = 0x8C
+const BME280_REGISTER_DIG_T3_MSB = 0x8D
+
+const BME280_REGISTER_DIG_P1_LSB = 0x8E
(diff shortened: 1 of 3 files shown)


dsockwell:
Oops, looks like I gave up trying to write tests. Something about having to fake multiple i2c reads. I'll see about finishing this.

deadprogram:
Thanks @dsockwell please let us know anything we can do to help.

deadprogram:
Hi @dsockwell any update here? We're able to provide assist if needed.

dsockwell:
Sure - the issue here is the test I wrote, platforms/i2c/bme280_driver_test.go , is  incomplete. Are meaningful tests strictly necessary for inclusion in gobot? It feels dirty, but I could null out the tests. This driver has been working continuously in my embedded system for weeks now.

Otherwise, the bme280 driver operates on several i2c transactions, whereas the test I copied from another driver only uses one. Is there another driver test that simulates multiple i2c read/write cycles that I could crib from?

Another option would be to modify bme280_driver.go to operate on the entire i2c output from the device for all i2c read access. The test could then provide the entire 255-byte read and have the driver read multiple times and sort it out.

Which of these ideas is most in line with how gobot is meant to work?

deadprogram:
Hi @dsockwell 

Can you please elaborate on what you mean by "operate on the entire i2c output from the device for all i2c read access"? I did not fully understand that, sorry.

Ideally you would have complete test coverage, but at the very least "build and smoke" tests are desirable. as far as multiple read/write, that sounds a a case where the function should probably be composed into smaller, more atomic functions for easier testing/readability.

Thanks!

deadprogram:
Hi @dsockwell if you were to submit a new PR with bme280_driver.go and bme280_driver_test.go under `/drivers/i2c` it would be gladly accepted. If you did not have time to do this, I'd be happy to help out. I have an BME280 around for testing.

platforms/i2c/bme280_driver.go
```diff
@@ -0,0 +1,433 @@
...
+	}
+	m.AddEvent(Error)
+	return m
+}
+
+func (h *BME280Driver) Name() string                 { return h.name }
+func (h *BME280Driver) Connection() gobot.Connection { return h.connection.(gobot.Connection) }
+
+// Start writes initialization bytes and reads from adaptor
+// using specified interval to accelerometer andtemperature data
```
elopio: This should be updated to mention humidity instead of accelerometer.

deadprogram:
I am working on an alternative implementation off a BME280 driver in this branch: https://github.com/hybridgroup/gobot/tree/feature/i2c/bme280

The only thing is currently is missing is adjusting the humidity reading based on the calibration settings.

dsockwell:
Will you include temperature and pressure readings in your driver, @deadprogram ?

deadprogram:
Notice that it has them by virtue of encapsulating the BMP180Driver: https://github.com/hybridgroup/gobot/blob/feature/i2c/bme280/drivers/i2c/bme280_driver.go#L37

The BME280 is a BMP180 that also has a humidity sensor.

dsockwell:
Ha! Good thinking, I like it.

deadprogram:
@dsockwell I saw you had already done some work to understand the calibration adjustment needed https://github.com/hybridgroup/gobot/pull/278/files#diff-ac932e15092d20aa6795db12d0bbf246R277

If you could help implement the `calculateHumidity()` method https://github.com/hybridgroup/gobot/blob/feature/i2c/bme280/drivers/i2c/bme280_driver.go#L108 it would be very much appreciated! :)

dsockwell:
I will give it a shot this morning.

dsockwell:
https://github.com/hybridgroup/gobot/pull/393

PR #289, opened 2016-06-14
support gpio pin turn on and off

ranjib:
I was doing [this tutorial](https://github.com/hybridgroup/gobot/pull/289), that uses Raspberry Pi to control a relay, but with gobot instead of python. And i felt it will be nice to have on and off for gpio pins in core gobot


ranjib:
@deadprogram any thoughts on this? 
I think go 1.2/1.3/1.4 related test failures are unrelated.

deadprogram:
Hi @ranjib golang 1.2/1,3 test fails are due to known incompatibility within the Paho MQTT libs. The 1.4 fail was due to intermittent problem with test timing for API tests that probably could use some work.

In any case, thanks for the PR. Now merging.

PR #291, opened 2016-07-13
platforms/bebop/client: fix type for binary.Write call

dgryski:
binary.Write only takes fixed-width types, of which 'bool' is not one.
Add a helper and turn those into int8s instead.


deadprogram:
Thanks, @dgryski merging.

PR #294, opened 2016-07-13
Move interface assertions to test files.

dgryski:
In general, these shouldn't live in the package proper, since they're
actually tests.


trevrosen:
Excellent! Really glad to see some work get done in this area. :shipit:

deadprogram:
Thanks, this is great. Merging.

dgryski:
My automated tooling only picked up some of these (and those are the ones I fixed), but a grep shows lots of other examples of this.  I'll file a PR for those too at some point.

PR #295, opened 2016-07-13
ble: populate descriptors after descovering characterisitcs

alytvynov:
On Linux, characteristic discovery requires setting the descriptor field
explicitly.


deadprogram:
Awesome!! Merging.

PR #296, opened 2016-07-13
Fix issues flagged by 'go vet'

dgryski:


deadprogram:
Looks good, thank you.

PR #297, opened 2016-07-17
Adding the initial NATS platform support

caledhwa:
This commit is to support NATS.io messaging protocol. This basic implementation supports the simplest use cases and follows directly the same implementation as the MQTT protocol. Parity is brought to bear between the two protocols. You can almost use them interchangeably except that the CLIENTID on the MQTT platform for GoBot uses a `string` while NATS uses an `int` for unique client. 

I intend to extend this code to add many more features from NATS. However, I wanted to at least get the platform docs, examples and happy-path use case implemented so that GoBotters (is that a thing?!?) could jump on using the NATS protocol for their projects. 

Finally, while I added some tests, once again I tried to meet the same type/number of tests that the MQTT platform added. This is weird to me because they added some tests which succeed only when the server is running and some that only succeed when the server is NOT running. I did meet both of these but don't really like this because there is no way to run the whole suite to success. I could fix this through mocking but with such little amount of code, I didn't want to expand the platform just for dependency injection and mocking - NOR did I want to add support for a mocking library and complicate the deployment process. I'm open to comments on this - I can work harder on the tests if needed. 

Thanks for consideration! (P.S. It was great meeting you @deadprogram at GopherCon. What a fantastic Hack Day! Best Ever! GO GOBOT!)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,7 @@ before_install:
  - cd gnatsd
  - "wget https://github.com/nats-io/gnatsd/releases/download/v0.8.1/gnatsd-v0.8.1-linux-amd64.zip"
  - unzip gnatsd-v0.8.1-linux-amd64.zip
+ - ls -la
  - ./gnatsd -p 4222 &
  - ./gnatsd -p 4223 --user test --pass testwd &
  - cd $HOME/gopath/src/github.com/hybridgroup/gobot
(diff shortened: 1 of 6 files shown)


caledhwa:
So, as I've seen I've broke the build (probably due to the unit tests), I'll refactor the tests. I need to see why the MQTT build WORKS but this one doesn't. Is there an MQTT server being used/set up during the testing?

caledhwa:
Finally, all checks have passed. I just needed to make some edits, add a few NATS servers and up the coverage. We have the first pass! Review away @deadprogram

deadprogram:
This looks great @caledhwa thank you so very much for your contribution! Merging.

caledhwa:
Hey! No problem @deadprogram - looking forward to adding more. What a great project. Thanks!

Issue #172, opened 2015-01-15
Servos: setting minimum and maximum "pulse width" - Arduino


pipejesus:
Using the Arduino libraries i can set the maximum and minimum pulse width for my servo so that i can calibrate it. Is it possible with Gobot ? What i mean is this : http://arduino.cc/en/Reference/ServoAttach

zankich:
@pipejesus at the moment no, we don't have that capability, however I'm currently working on a refactor of our firmata implementation and based off of the [servo](https://github.com/firmata/protocol/blob/master/servos.md) protocol document, it should be easy to add.

pipejesus:
Superb! Great project, thank you Adrian for bringing Gobot to Earth :-)

losinggeneration:
It looks like this might be implemented (at least on the client library.) https://github.com/hybridgroup/gobot/blob/dev/platforms/firmata/client/client.go#L217 However, it looks like the bytes are sent as max,min instead of min,max unless I'm misunderstanding the code and/or protocol doc.

deadprogram:
Hi @losinggeneration looks to me like you are absolutely correct. Would you be willing to submit a PR with that fix?

losinggeneration:
I'll pick up a servo tomorrow to verify it works and then get a PR in for it.

deadprogram:
Seems like this was addressed already. Please reopen if needed.

PR #300, opened 2016-07-21
Fix the ServoConfig byte order

losinggeneration:
This corrects the byte order according to the [servo](https://github.com/firmata/protocol/blob/master/servos.md) protocol specification. It describes the bytes to be sent as:

```
// minPulse and maxPulse are 14-bit unsigned integers
0  START_SYSEX          (0xF0)
1  SERVO_CONFIG         (0x70)
2  pin number           (0-127)
3  minPulse LSB         (0-6)
4  minPulse MSB         (0-13)
5  maxPulse LSB         (0-6)
6  maxPulse MSB         (7-13)
7  END_SYSEX            (0xF7)
```

While this seems to fix the Firmata client, issue #172 seems to be asking for this to be exposed in the FirmataAdapter. It would be easy enough to add to the adapter, and if that's wanted I can update the PR to include that.

### DIFF
--- a/platforms/firmata/client/client.go
+++ b/platforms/firmata/client/client.go
@@ -218,10 +218,10 @@ func (b *Client) ServoConfig(pin int, max int, min int) error {
 	ret := []byte{
 		ServoConfig,
 		byte(pin),
-		byte(max & 0x7F),
-		byte((max >> 7) & 0x7F),
 		byte(min & 0x7F),
 		byte((min >> 7) & 0x7F),
+		byte(max & 0x7F),
+		byte((max >> 7) & 0x7F),
 	}
 	return b.writeSysex(ret)
 }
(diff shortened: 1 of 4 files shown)


deadprogram:
Hi @losinggeneration thanks for the PR. Yes, please, can you also add to the Gobot adaptor?

losinggeneration:
Yes. I'm also currently writing unit tests for ServoConfig. After I'm finished with that. I'll add the adapter.

losinggeneration:
I've added this to the FirmwareAdapter. If this is something all platforms could benefit from, adding it to the gpio.ServoDriver would be a possibility. However, that will require that all platforms be updated to implement ServoConfig.

losinggeneration:
I just realized I didn't sign off on all commits. Should I rebase to correct this?

deadprogram:
Hi @losinggeneration ideally yes, but if not, this thread should serve as indication of your intent to contribute :) Let me know if you end up doing so, and thanks again.

losinggeneration:
Sorry for the delayed response. I've updated the commits with sign offs. I'd there's anything else needed let me know.

deadprogram:
Thanks again, @losinggeneration merging.

Issue #302, opened 2016-08-18
Is it possible to use gobot libraries without all the gobot/robot initialization. 


ericho:
I'm wondering if it's possible to have a program like this 

```go
package main

import (
    "time"

    "github.com/hybridgroup/gobot/platforms/gpio"
    "github.com/hybridgroup/gobot/platforms/intel-iot/edison"
)

func main() {
    e := edison.NewEdisonAdaptor("edison")
    led := gpio.NewLedDriver(e, "led", "13")
    for {
        led.Toggle()
        time.Sleep(1)
    }
}
```

That could be part of another program. For instance if I want to read values from a sensor and send it to a database or another process. I think that I can achieve this by using a channel but the idea would be to have a simple way to get the ports/sensors data, something like the mraa library. 

Is there a way to have such functionality on gobot? 

Thanks

trevrosen:
Try something like this, but with the Edison adaptor and let me know what happens.

https://github.com/hybridgroup/gobot/blob/dev/examples/beaglebone_basic_direct_pin.go

trevrosen:
You're basically on the right track, you just need to `Connect()` the adaptor.

ericho:
Awesome! The `connect()` was the thing I needed. Now this code works

```go
package main

import (
	"time"
	"github.com/hybridgroup/gobot/platforms/gpio"
	"github.com/hybridgroup/gobot/platforms/intel-iot/edison"
)

func main() {
	e := edison.NewEdisonAdaptor("edison")
	led := gpio.NewLedDriver(e, "led", "13")
	e.Connect()
	for {
		led.Toggle()
		time.Sleep(1000 * time.Millisecond)
	}
}
```

Thanks :)

deadprogram:
One other small thing worth mentioning. You should also call `Start()` on the driver, to ensure that it is properly initialized:

```
package main

import (
    "time"
    "github.com/hybridgroup/gobot/platforms/gpio"
    "github.com/hybridgroup/gobot/platforms/intel-iot/edison"
)

func main() {
    e := edison.NewEdisonAdaptor("edison")
    led := gpio.NewLedDriver(e, "led", "13")
    e.Connect()
    led.Start()
    for {
        led.Toggle()
        time.Sleep(1000 * time.Millisecond)
    }
}
```

deadprogram:
We should update the docs with this info, so people realize they can use Gobot as a set of Golang device libraries, not just use the entire framework.

[...]

PR #303, opened 2016-08-24
Add example for Edison blink demo without gobot initialization.

ericho:
This example shows how to use gobot as another golang library.
Issue #302 
Signed-off-by: Erich Cordoba <erich.cm@yandex.com>


deadprogram:
Thanks, @ericho

PR #306, opened 2016-08-31
Refactor eventer to use event channels & handlers

deadprogram:
This PR changes how Gobot internally handles events. It rewrites the Eventer to include all of the relevant functionality, and uses channels & selects to manage the event handling. The  `On()` and `Once()` functions have been moved into the Eventer interface, and use the event channels as their implementation. 

It is important to note that these functions are now called off of the driver/adaptor that implement them, no longer off of the global `gobot` package. This is a breaking change to the current interface.

In this example, the `On()` functions are called off of `button`:
```go
...
	button := gpio.NewButtonDriver(e, "myButton", "2")
	led := gpio.NewLedDriver(e, "myLed", "7")

	work := func() {
		button.On(gpio.ButtonPush, func(data interface{}) {
			led.On()
		})
		button.On(gpio.ButtonRelease, func(data interface{}) {
			led.Off()
		})
	}
...
```

The above code is the same as doing the following code "metal" using Gobot as a library not a full framework, by writing code to use just the event channel along with a `select()`:

```go
...
	led := gpio.NewLedDriver(e, "led", "13")
	button := gpio.NewButtonDriver(e, "button", "5")

	led.Start()
	button.Start()

	buttonEventChannel := button.Subscribe()
	for {
		select {
		case event := <-buttonEventChannel:
			if event.Name == gpio.ButtonPush {
				led.On()
			}
			if event.Name == gpio.ButtonRelease {
				led.Off()
			}
		}
	}
}
```

If you use the new `On()` function, it automatically runs your handler inside a goroutine. The above code is just a simple example, but could very easily do the same thing.


trevrosen:
@deadprogram this is a monster PR -- any guidance on how I should break it down for review?

deadprogram:
@trevrosen indeed, it is rather sizable, which is why I've asked the community for help :)

The changes here can be thought of as 3 sets of related changes:

- Gobot core changes to Eventer/Event and utils
- Changes to various drivers/adaptors to accommodate the new Eventer
- Changes to the examples that consume drivers/adaptors

Hopefully that provides a little more clarity as to the nature of the changes in this PR. Thanks again for your help!

trevrosen:
@deadprogram awesome I'll get into it.

eventer.go
```diff
@@ -1,36 +1,116 @@
...
+// handles events.
 type Eventer interface {
-	// Events returns the Event map.
-	Events() (events map[string]*Event)
-	// Event returns an Event by name. Returns nil if the Event is not found.
-	Event(name string) (event *Event)
-	// AddEvent adds a new Event given a name.
+	// Events returns the map of valid Event names.
+	Events() (eventnames map[string]string)
+	// Event returns the map of valid Event names.
```
trevrosen: Doc here seems incorrect, since `Event` returns a string

Also wondering if maybe `Event` shouldn't be `GetEvent` or something to match up as an obvious companion with `AddEvent`.

eventer.go
```diff
@@ -1,36 +1,116 @@
...
+
+	// slice of out channels used by subscribers
+	outs []eventChannel
 }
 
-// Eventer is the interface which describes behaviour for a Driver or Adaptor
-// which uses events.
+// Eventer is the interface which describes how a Driver or Adaptor
+// handles events.
 type Eventer interface {
```
trevrosen: *Caveat for all this feedback: I need to get all my Gobot knowledge back into a higher-lvl mental cache than it's in right now, so please forgive me if some of this comes across as a bit out-of-date or ignorant.*

If I want to delete an event, it would seem that right now I would just call `AddEvent` in such a way as to force a key collision and reset the value in the map. I suggest having `DeleteEvent` for removing a key and to make `AddEvent` throw an error if you try to add one with the same name. Does that make sense for what you're trying to achieve with this interface?

deadprogram: Yes, makes perfect sense. Thanks.

eventer.go
deadprogram: That is probably true, so as to clarify the exact intentions here. Thanks.

trevrosen:
@deadprogram I'm still grokking/reviewing, but I like the change so far. I hope that we can come up with a good narrative-style documentation for this so that it gets quick adoption!

eventer.go
deadprogram: Actually, I think we should leave it just `Event()` so we can keep the nice looking syntax like:

```go
sensor.On(sensor.Event("data"), func(data interface{})
```

I did update the comments as you suggested @trevrosen

rafael-azevedo:
The documentation for On() Off() functions need to be changed universally. I spent a good amount of time troubleshooting today when the issue was this change all along .

deadprogram:
Hi @rafael-azevedo you are correct, sorry about the confusion. The repo should have all been updated, however the site still shows older syntax.

We did not mean to release all this new functionality until the upcoming 1.0 version. However, I misunderstood how Golang was getting the code from Github, as discussed here https://github.com/hybridgroup/gobot/issues/319#issuecomment-252676389

Very sorry about the confusion, and the upcoming release should have all the code + docs in sync. If you want to see what is coming next, please checkout the `dev` branch: https://github.com/hybridgroup/gobot/tree/dev

rafael-azevedo:
@deadprogram  thanks for the info. I love the project , just trying to make sure there was not looked over .

PR #309, opened 2016-09-04
Remove debug message from i2c_device.go

ericho:
Signed-off-by: Erich Cordoba <erich.cm@yandex.com>


deadprogram:
Thanks @ericho for the correction.

Issue #262, opened 2016-02-18
Errors on Intel Edison, Sans-Arduino


ryanhatfield:
I have an Intel Edison without the Arduino Development board, and the [edison_adapter](https://github.com/hybridgroup/gobot/blob/dev/platforms/intel-iot/edison/edison_adaptor.go#L199) is looking for pins that don't exist.

According to this [Guide on Multiplexing](http://www.emutexlabs.com/project/215-intel-edison-gpio-pin-multiplexing-guide) with the Intel Edison, some of the pins with hard references in code won't exist if you're using the Intel Edison without the pin expanders.

Removing any [GPIO Pins larger than 200](https://github.com/hybridgroup/gobot/blob/dev/platforms/intel-iot/edison/edison_adaptor_test.go#L64) should help show what would happen if those pins are missing, and any pins over 200 are added by the i2c port expander. (as I understand it).

I'll be working to remove the pins that are causing issues right now, but is there any larger issues I'll run into with these pins being unavailable?

deadprogram:
Hi, @ryanhatfield thanks for pointing this out, you are correct that the current implementation makes some assumptions about the Arduino-compatible breakout board being present. 

Are you using the mini-breakout board, the Sparkfun board, or something else entirely?

One possible solution that comes to mind, is pulling out the shared Edison muxing functions into a separate file, and then creating new adaptors for each of the specific breakout boards.

What do you think?

ryanhatfield:
I think having the _Arduino_ section as it's own _Platform_ or device would be another way to solve this. I would think since the i2c expansion is being added, a "i2c port expander" device could be added, and the edison arduino might have a platform specific version to handle the unique way it gets added?

I have the mini-breakout board, the Sparkfun Edison blocks, and a fully custom board at my disposal, so pure abstraction of the Intel Edison's base functionality is a requirement for me. 

Also, I do a certain amount of Kernel configuration for some pins / registers, I will likely look into a way of dynamically or statically defining the pin mappings (if different than default) and that could also facilitate the additional pins available on the Arduino variant.

ryanhatfield:
Also, with the speed I'm moving at with some of these things, I might not take the time to back port functionality into all platforms right away. I might just keep a branch open on my fork to hold these modifications and when someone (possibly me) has time, they can be fully integrated or modified to be pulled back upstream.

deadprogram:
[...]

PR #318, opened 2016-10-07
Added return as arduinoSetup was always being called

devfubar:
After making this change I managed to get the blinking LED example working on a Sparkfun Edison with GPIO block.  This helps with issue #262


deadprogram:
Awesome! This is great news.

Issue #317, opened 2016-10-06
Edison documentation


devfubar:
I am new to gobot and I was trying to get my Intel Edison working. I do not have the arduino breakout board and spent some time figuring our that you make an assumption that it is present when using the Edison. I am interested in issue #262 however I think that the documentation should be updated to state the Arduino breakout board assumption to prevent people scratching their heads for as long as I did.

deadprogram:
Hi @devfubar that is no doubt true. Once the current refactoring in PR has been merged into master, I expect to add some code to handle the Arduino as well as the mini-breakout board, as well as to show other contributors how to more easily add other Edison breakout boards.

devfubar:
Thanks @deadprogram for getting back to me.  I was just interest to see if this was still the plan as the last comment on #262 was in Feb. I raised this issue as an interim solution to prevent people from wondering why the basic example wasn't working.

I'm happy to contribute so some sort of guide that you mentioned would be very useful.

deadprogram:
The `dev` branch now has all of these changes, thank you very much @devfubar for your awesome contributions on this.

PR #326, opened 2016-10-22
Pinmap for Intel Edison Sparkfun and documentation

devfubar:
This should fix issues #262 and #317


platforms/intel-iot/edison/sparkfun_pinmap.go
```diff
@@ -0,0 +1,116 @@
+package edison
+
+var sparkfunPinMap = map[string]sysfsPin{
+  "12": sysfsPin{
```
deadprogram: Shouldn't this be labeled "GP12" to exactly match the labels on the board? Likewise with the reset of the pins.

devfubar: I did think about that as it is what is on the board it self.  However most of the documentation including a lot of the Intel stuff references just the port number eg `echo 14 > /sys/class/gpio/export`.  When I came to try the code for the first time my natural instinct was to use "14" so thats what swayed it in my mind.....

deadprogram: I get what you are saying, seems like a good thing. 

At some point I think I will implement something to strip off unneeded but nice looking alpha prefixes, just to match board labeling, but that is entirely separate.

deadprogram:
Great stuff, @devfubar thank you.

devfubar:
No problem @deadprogram happy to help especially since you guys are saving em a lot of time :-)  I may add a few sensors and such as I go through, currently working on srf05 as I have a few of them kicking about...

PR #330, opened 2016-10-27
Adds support for Grove Magnetic Switch

wfernandes:
I didn't add a test because it uses the underlying button driver which was tested. 

However, I did run a physical test by flashing code to an Arduino UNO and verifying behavior.

I'm not sure why the version got updated in `version.go` as it wasn't part of my commit.


wfernandes:
Nevermind. I figured out what I did. Closing until I fix the issue.

PR #331, opened 2016-10-27
Adds support for Grove Magnetic Switch Sensor

wfernandes:
- http://wiki.seeed.cc/Grove-Magnetic_Switch/

I didn't add a test because it uses the underlying button driver which was tested.
However, I did run a physical test by flashing code to an Arduino UNO and verifying behavior.


deadprogram:
Thanks, @wfernandes

PR #332, opened 2016-10-27
Minor improvements

wfernandes:
- Uses time.NewTimer() instead of time.After() since it is more efficient to just use one timer rather than creating a new Timer every time in the tight loop.
- Fixes tests and makes them relatively consistent.
- Added some other useful assertions to existing tests.


wfernandes:
The test that failed seems to be completely irrelevant to the changes made. I've ran the `firmata/client/client_test.go` tests locally and they pass for me. Can you confirm?

deadprogram:
This is very good stuff @wfernandes thanks very much. Makes me want to get rid of even more `time.After()` calls now.

PR #334, opened 2016-11-06
Adds tests for grove drivers

wfernandes:

- Tried to write the tests without depending on un-exported properties like the `halt` channel.

### DIFF
--- a/drivers/gpio/grove_drivers_test.go
+++ b/drivers/gpio/grove_drivers_test.go
@@ -39,11 +39,8 @@ func TestDriverDefaults(t *testing.T) {
 	}
 
 	for _, driver := range drivers {
-		t.Run(getType(driver), func(t *testing.T) {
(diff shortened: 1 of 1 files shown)


wfernandes:
Just noticed that the build failed because `t.Run()` is not available in go1.6 

If support for go1.6 is required I can update the tests to not use [Subtests feature of 1.7](https://blog.golang.org/subtests)

deadprogram:
Hi @wfernandes thanks for working on this. Currently we're trying to always target the latest Golang release, as well as maintain backwards compatibility with the release immediately before the current one. Hence the dev build right now uses https://github.com/hybridgroup/gobot/blob/dev/.travis.yml#L6-L8

If you would not mind making the change your suggest for golang 1.6+ I think many users would appreciate it, as would I. Thank you!

wfernandes:
@deadprogram Sure. Will make the changes.

deadprogram:
Thanks, @wfernandes merging.

PR #337, opened 2016-11-13
Ranges over channel instead of using select

wfernandes:


deadprogram:
Thanks @wfernandes for this. I just merged into dev.

PR #338, opened 2016-11-14
Add GoPiGo Platform

JohannWeging:
This is a WIP pull request to get feedback and help on the design. 
Furthermore this is the first time I'm working with Gobot and the GoPiGo platform, so if some of the decisions I make seem totally bonkers feel free to correct me.  

The implementation is based of the Python and C implementation from [Dexter Industries](https://github.com/DexterInd/GoPiGo/tree/master/Software).

Some questions regarding the design:

The `NewGoPiPoAdaptor` returns an error because it starts the I2C communication between the Rasberry Pi and the GoPiGo board. Can it stay this way or should the start call to I2C be moved into its own method?  

At the moment the commands to control the robot are tied to the platform. Should this be moved 
into one or multiple drivers?

I implemented the `DigitalRead` Method as an example for the Gobot interface. The pin gets translated with the help of the `pins` map. The pin numbers are limited because of the way its done in the [Python implementation](https://github.com/DexterInd/GoPiGo/blob/master/Software/Python/gopigo.py#L298).
And returns `erros.New("no such pin")` if a wrong pin number is provided.
@DexterInd can you tell me why it's limited this way in the read methods but in the wirte methods its commented out?  

The I2C communication uses it's own buffer for every call. I could use one global buffer to reduce 
the GC noise but then I would need to take care of concurrent API calls? 

The next question is about how much work should be offloaded into the Adaptor.
One example: 
To turn the robot 90° left it's necessary to do the Encoder calculations for movement manually:
```go
gopigo.EncoderTarget(false, true, 90/360.0/64)
gopigo.TurnLeft()
```
I cloud change the movement methods to:
```go
// 0 disables the encoder
func (g *GoPiGoAdaptor) TurnLeft(degrees int) 
```
Or:
```
// passing no value disables the encoder
func (g *GoPiGoAdaptor) TurnLeft(degrees ...int)
```
This would affect methods like `Forward` where it would be possible specify a distance in cm.

### DIFF
--- a/drivers/i2c/gopipo_driver.go
+++ b/drivers/i2c/gopipo_driver.go
@@ -1,0 +1,262 @@
+package i2c
+
+import (
+	"time"
+
+	"github.com/hybridgroup/gobot"
+)
+
+// debug=0
+//
+// us_cmd				=[117]		#Read the distance from the ultrasonic sensor
+// en_com_timeout_cmd	=[80]		#Enable communication timeout
+// dis_com_timeout_cmd	=[81]		#Disable communication timeout
+// timeout_status_cmd	=[82]		#Read the timeout status
+// enc_read_cmd		=[53]		#Read encoder values
+// trim_test_cmd		=[30]		#Test the trim values
+// trim_write_cmd		=[31]		#Write the trim values
+// trim_read_cmd		=[32]
+//
+// ir_read_cmd			=[21]
+// ir_recv_pin_cmd		=[22]
+// cpu_speed_cmd		=[25]
+
+// v16_thresh=790
+const (
+	GOPIGO_ADDRESS = 0x08 // i2c address for the GoPiGo
+
+	GOPIGO_STOP             = 0x78 // stop the gopigo
+	GOPIGO_FIRMWARE_VERSION = 0x10 // get the version of the GoPiGo
+
+	GOPIGO_MOTOR1           = 0x6F // control motor 1
+	GOPIGO_MOTOR2           = 0x70 // control motor 2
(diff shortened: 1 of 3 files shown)


JohannWeging:
I split up the GoPiGo implementation into a platform and an I2C driver. I tried to make the driver easy to use while keeping the platform compatible with existing gobot components. 
There are still some functions missing and no unit tests. 

The 'gopigo_driver_test.go' has an example function using the GoPiGo platform with the GoPiGo driver and the gobot LED driver.

deadprogram:
Hi @JohannWeging first of all, thanks for working on this.

I would move the driver you wrote to under `/platforms/gopigo` since it only applies to the GoPiGo. Things under `/drivers` are those intended to be able to work on on multiple platforms.

Regarding your proposed `TurnLeft()` API, I like your proposed version more than the original one you describe, certainly.

Lastly, please make sure that your tests all pass locally. Once you move the driver into the "correct" location, it might be easier to look into why the tests are not passing.

Looking good, please keep going!

trevrosen:
Hi @JohannWeging are you still working on this? Happy to help you get it over the goal line if you're stuck. 😸

eduncan911:
Same here.

trevrosen:
This PR will have a one-year anniversary next month. I humbly suggest it should be closed if no activity by then. 😸

eduncan911:
This PR seems to have been superseded by #438 .

And then other devs jumped in and made a final PR #448 which finally got merged.

So yes, this one is dead.

trevrosen:
@deadprogram can we close this out? Selfishly I just want it out of my Trailer.app list... 😈

deadprogram:
OK, it is officially muerto. Thanks everyone.

PR #340, opened 2016-11-21
Adds note about the kernel version for CHIP

wfernandes:
- The current ChipAdaptor will only support OS 4.3 because of the
GPIO pin mapping.

See http://docs.getchip.com/chip.html#gpio for more details.


deadprogram:
Hi @wfernandes do you think we should possibly support the newer CHIP update, instead of this? Or possibly support both?

wfernandes:
Hi Ron. I think we should support both.
I was planning on submitting a PR for it after the Thanksgiving holidays which would also include usage docs for either platform.

deadprogram:
Hi, @wfernandes check out the branch here, it should work with either version of the OS:

https://github.com/hybridgroup/gobot/tree/feature/chip-44

PR #341, opened 2016-11-30
Support for C.H.I.P. 4.4

deadprogram:
Adds support for the new C.H.I.P. OS v4.4 alongside version 4.3 support. Nothing needed from the developer, the code should auto-recognize which OS is running and make suitable adjustments.

### DIFF
--- a/commander.go
+++ b/commander.go
@@ -22,15 +22,18 @@ func NewCommander() Commander {
 	}
 }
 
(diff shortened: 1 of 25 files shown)


deadprogram:
I've verified on both 4.3 and 4.4, so merging into dev. Thanks everyone!

PR #342, opened 2016-11-30
Support for newer Beaglebone 4.x kernels

deadprogram:
Adds support for the newer Beaglebone v4.x kernels alongside older version 3.x support. Nothing needed from the developer, the code should auto-recognize which OS is running and make suitable adjustments.


nevers:
Great work! Sadly, I won't have the time to test it any time soon.

PR #344, opened 2016-11-30
platforms/firmata/client: use io.ReadFull

### LINKED ISSUE #343: platforms/firmata/client: invalid read
hi,

I am reading data from an analog device connected to pin-A1.
I was reading weird values:

```
2016/11/30 21:42:11 ---analog-read-1: 856 [225 88 6]
2016/11/30 21:42:11 ---analog-read-1: 856 [225 88 6]
2016/11/30 21:42:11 ---analog-read-1: 0 [225 0 0 88 6]
2016/11/30 21:42:11 ---analog-read-1: 857 [225 89 6]
2016/11/30 21:42:11 ---analog-read-1: 857 [225 89 6]
2016/11/30 21:42:11 ---analog-read-1: 0 [225 0 0 89 6]
2016/11/30 21:42:11 ---analog-read-1: 857 [225 89 6]
2016/11/30 21:42:11 ---analog-read-1: 856 [225 88 6]
2016/11/30 21:42:11 ---analog-read-1: 0 [225 0 0 87 6]
```
ie: first value is the `int` value from the 2 bytes in the buffer (_ie_: `[255 88 6]` in the first line) after the messageType byte "flag".

in the log above, one can see that for the 2 first values, I do read a correct value, with a buffer of 3 bytes (which isn't surprising as this created from a call to `b.read(3)`.)

unfortunately, there were a few spurious cases where the output buffer size from `b.read(3)` wasn't `3`.
(_eg_ at the 3rd line and the last line of the log)

it's an issue in `b.read` which should use `io.ReadFull(b.connection, buf[:])`.

I'll send a PR.

sbinet:
This CL uses io.ReadFull to make sure exactly n bytes are read, and
correctly handle io.EOF and co.

Fixes #343.

### DIFF
--- a/platforms/firmata/client/client.go
+++ b/platforms/firmata/client/client.go
@@ -315,21 +315,9 @@ func (b *Client) write(data []byte) (err error) {
 	return
 }
 
-func (b *Client) read(length int) (buf []byte, err error) {
-	i := 0
-	for length > 0 {
-		tmp := make([]byte, length)
-		if i, err = b.connection.Read(tmp); err != nil {
-			if err.Error() != "EOF" {
-				return
-			}
-			<-time.After(5 * time.Millisecond)
-		}
-		if i > 0 {
-			buf = append(buf, tmp...)
-			length = length - i
-		}
-	}
+func (b *Client) read(n int) (buf []byte, err error) {
+	buf = make([]byte, n)
+	_, err = io.ReadFull(b.connection, buf)
 	return
 }
 


sbinet:
not sure if this considered as hot-fix or not (thus the PR is based off `master`)

deadprogram:
Hi @sbinet thanks for this. I will merge into master, and also take care of cherry-picking this commit into the dev branch.

PR #350, opened 2016-12-29
Add the I2C driver for BMP180

elopio:


elopio:
This is my first attempt at playing with i2c, so any feedback will be appreciated.

deadprogram:
Hi, @elopio this is great, thank you. I have a couple of comments/requests.

- Can you please resubmit against the `dev` branch instead of `master`. This allows us to have a proper release of new features. due to how golang fetches packages.
- For i2c drivers, instead of a goroutine to update the data, the typical pattern has been to let the user call at whatever periodicity they wish. The HMC6352 is a good example of this. The `Start()` method just initializes the i2c commication and does whatever configuration is needed. In your case, that would mean removing lines 113-127 of bmp180_driver.go

Thanks again for working on this!

elopio:
on it.

elopio:
@deadprogram thanks for the review. Please check it again.
I moved the mode as an argument to the Pressure method. I'm not sure if I should have left it as an attribute of the driver.

deadprogram:
Thank you very much for this, @elopio now merging.

elopio:
\o/ thank you!

PR #351, opened 2016-12-30
Add the I2C driver for L3GD20H gyroscope

elopio:
https://www.adafruit.com/products/1032

### DIFF
--- a/drivers/i2c/l3gd20h_driver.go
+++ b/drivers/i2c/l3gd20h_driver.go
@@ -1,0 +1,136 @@
+package i2c
+
+import (
+	"bytes"
+	"encoding/binary"
+	
+	"gobot.io/x/gobot"
+)
+
+const l3gd20hAddress = 0x6B
+
+// Control Register 1
+const l3gd20hRegisterCtl1 = 0x20
+const l3gd20hNormalMode = 0x8
+const l3gd20hEnableZ = 0x04
+const l3gd20hEnableY = 0x02
+const l3gd20hEnableX = 0x01
+
+// Control Register 4
+const l3gd20hRegisterCtl4 = 0x23
+
+const l3gd20hRegisterOutXLSB = 0x28 | 0x80 // set auto-increment bit.
+
+// L3GD20HDriver is the gobot driver for the Adafruit Triple-Axis Gyroscope L3GD20H.
+// Device datasheet: http://www.st.com/internet/com/TECHNICAL_RESOURCES/TECHNICAL_LITERATURE/DATASHEET/DM00036465.pdf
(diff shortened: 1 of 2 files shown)


elopio:
I had some fun today with this gyroscope. But there are quite a few things that I don't know how to write for gobot. In addition to scale, this has high-pass filter, different fifo queue modes, interrupts, output data rate and bandwidth selection. I don't really understand how to use all of them. I usually look at the adafruit library when I get lost, but they are just using the default values for all of that.
It would be just fine to use the default values in here, but go doesn't have optional parameters. So if we ever want to support something else, it will break the API.

Please take a look.

I would also be OK donating this PR to somebody who know more about this gyroscope :)

deadprogram:
Hi @elopio seriously love all you are doing! Can you please change this PR over to the `dev` branch? Thanks for your efforts!

elopio:
Damn it! I'm never going to get it right. I changed the base, sorry about that.

drivers/i2c/l3gd20h_driver.go
```diff
@@ -0,0 +1,135 @@
...
+	// L3GD20HScale2000dps is the 2000 degress-per-second scale.
+	L3GD20HScale2000dps = 0x30
+)
+
+// NewL3GD20HDriver creates a new driver with the i2c interface for the L3GD20H device.
+func NewL3GD20HDriver(c I2c, s L3GD20HScale) *L3GD20HDriver {
+	return &L3GD20HDriver{
+		name:       "L3GD20H",
+		connection: c,
+		scale: s,
```
deadprogram: I suggest instead of requiring a param for scale, instead set a sensible default in line 49, and then add a `SetScale()` method. What do you think?

elopio: That sounds good. I have made the change.

I was thinking a little about this, and it seems that it's a common thing for most drivers to have some sort of options. Maybe for later, when you bump the major version, all drivers could take in the init a kind of generic options as a parameter, and offer a SetOptions in all of them to change it.

deadprogram:
Thanks, @elopio this looks great. Now merging... :globe_with_meridians:

elopio:
Thanks @deadprogram. I also have here an accelerometer: https://www.adafruit.com/products/163
But if I'm understanding correctly, that would be just 3 analog sensors. Should I make an driver for it?

deadprogram:
Hi @elopio I think an analog accelerometer driver that handled the scaling from a raw analog value to a specific Gs value is useful. It should also require 3 analog pins, somewhat like how the `RGBLEDDriver` requires 3 PWM pins.

PR #355, opened 2017-01-11
Fix for GPIO on C.H.I.P running 4.4.13 kernel

erkkah:
This is a fix for making GPIO work on all different kernel versions.
Before this fix, GPIO fails silently on kernel versions after 4.4.11.
It changes the hard coded implementation to a dynamic approach.
And.. I added definitions for all available GPIO pins.

Not sure if this counts as a hot fix, but C.H.I.P GPIO is broken without the fix.


deadprogram:
Hi @erkkah thanks for this PR, I like the more dynamic approach. Also, like the additional pin mappings. 

Could you please resubmit this to the `dev` branch? Thank you.

erkkah:
Ok, great. Will do!

PR #356, opened 2017-01-12
Fix for GPIO on C.H.I.P running 4.4.13 kernel (now on the dev branch)

erkkah:
This is a fix for making GPIO work on all different kernel versions.
Before this fix, GPIO fails silently on kernel versions after 4.4.11.
It changes the hard coded implementation to a dynamic approach.
And.. I added definitions for all available GPIO pins.


deadprogram:
Thanks @erkkah now merging.

Issue #247, opened 2015-11-20
Restrictive I2c modelization


abourget:
Hi there,

I'm trying to implement the Edison miniboard and I see that the I2cStart, I2cRead and I2cWrite functions are all attached to the Adaptor. However, the Edison miniboard has 2 I2C controllers..

The mraa exposes the different I2C controllers using an `i2c_context`.. and you can have many such contexts (or `devices`) for a given board.

Also, it's not possible for a user to select a certain I2C device, the Adaptor has its default and that's it.

Could you imagine reworking this design ?  Would such a proposition be mergeable ?

abourget:
I'm looking at many things in `mraa` that we could use.. is the goal here to have a clean rewrite of those features ? or could we imagine using `mraa` (cgo)  at some point ?

deadprogram:
You could be easily done in Gobot by extending via the needed Linux files paths to the other i2c buses in the intel-iot-/edison platform. I do not think you need to drag in all of MRAA just for that.

MRAA has the limitations of blocking, and cgo also has many limitations. The current implementation has the advantage of being pure Go.

That said, there are platforms in Gobot that use cgo wrappers, such as opencv. Nothing prevents a parallel implementation that wraps mraa and/or upm. But the pure Golang approach has many benefits.

abourget:
Yeah.. I'd want to keep pure go too.. what about adding a layer behind I2cStart / I2cRead / I2cWrite .. which would use a `DefaultI2cBus`.. and add the notion of an I2cBus .. I2cStart for all platforms that have only one would continue to work, creating a Adaptor.I2cBus .. but some adaptors could expose the different `I2cBus`es as a list or map or something..

This would keep retrocompatibility, keep the easy way like it is, but allow for adaptors with many buses.. and give you a way to pass the I2cBus around if you want..

What do you think ?

deadprogram:
That sounds like a good approach towards a first implementation to me. If you work on this, and you want feedback, submit a PR early with comment that it is Work In Progress and not ready to merge. Or just do it! :)

abourget:
I wanted to implement the LSM9DS0 first.. and see what would be needed.

You can take a look here: https://github.com/abourget/gobot/commits/master

The i2c interface isn't very slick.. but I do feel it should be more furnished.. It *is* a re-implementation
[...]

PR #358, opened 2017-01-22
Change to make I2C support more than one bus and other access methods than "block"

erkkah:
Hi there!

These are my suggested changes to the I2C interface.

Based on the changes discussed in https://github.com/hybridgroup/gobot/issues/247,
I started out with @abourget's changes to sysfs, and added a new layer in drivers/i2c
to support multiple buses.

The old interface is kept in parallel, I've only implemented the required platform
changes in the c.h.i.p adaptor.

For adaptors, there is a new interface, which replaces the I2c interface: 

```
type I2cConnector interface {
	I2cGetConnection(address int, bus int) (device I2cConnection, err error)
}
```

Driver implementations will then use this interface to get a connection to a
device at a given address + bus combination.

I've also added more fine grained I2C methods than just block read / write,
which simplifies driver implementation.

See https://github.com/erkkah/gobot/blob/tsl2561-driver/drivers/i2c/tsl2561_driver.go
for a driver example using the new interface.

### DIFF
--- a/drivers/i2c/helpers_test.go
+++ b/drivers/i2c/helpers_test.go
@@ -17,21 +17,74 @@ var blue = castColor("blue")
 type i2cTestAdaptor struct {
 	name         string
 	written      []byte
+	pos          int
 	i2cReadImpl  func() ([]byte, error)
 	i2cWriteImpl func() error
 	i2cStartImpl func() error
 }
 
 func (t *i2cTestAdaptor) I2cStart(int) (err error) {
+	t.pos = 0
 	return t.i2cStartImpl()
 }
+
 func (t *i2cTestAdaptor) I2cRead(int, int) (data []byte, err error) {
 	return t.i2cReadImpl()
 }
+
 func (t *i2cTestAdaptor) I2cWrite(address int, b []byte) (err error) {
 	t.written = append(t.written, b...)
 	return t.i2cWriteImpl()
 }
+
+func (t *i2cTestAdaptor) ReadByte() (val uint8, err error) {
+	bytes, err := t.i2cReadImpl()
+	val = bytes[t.pos]
+	t.pos++
+	return
+}
+
+func (t *i2cTestAdaptor) ReadByteData(reg uint8) (val uint8, err error) {
+	bytes, err := t.i2cReadImpl()
+	return bytes[reg], err
+}
+
+func (t *i2cTestAdaptor) ReadWordData(reg uint8) (val uint16, err error) {
+	bytes, err := t.i2cReadImpl()
+	low, high := bytes[reg], bytes[reg + 1]
+	return (uint16(high) << 8) | uint16(low), err
+}
+
+func (t *i2cTestAdaptor) ReadBlockData(b []byte) (n int, err error) {
+	reg := b[0]
+	bytes, err := t.i2cReadImpl()
+	copy(b, bytes[reg:])
+	return len(b), err
+}
+
+func (t *i2cTestAdaptor) WriteByte(val uint8) (err error) {
+	t.pos = int(val)
+	t.written = append(t.written, val)
+	return t.i2cWriteImpl()
+}
+
+func (t *i2cTestAdaptor) WriteByteData(reg uint8, val uint8) (err error) {
+	t.pos = int(reg)
+	t.written = append(t.written, reg)
+	t.written = append(t.written, val)
+	return t.i2cWriteImpl()
+}
+
+func (t *i2cTestAdaptor) WriteBlockData(b []byte) (err error) {
+	t.pos = int(b[0])
+	t.written = append(t.written, b...)
+	return t.i2cWriteImpl()
+}
+
+func (t *i2cTestAdaptor) I2cGetConnection( /* address */ int /* bus */, int) (connection I2cConnection, err error) {
+	return t, nil
+}
+
 func (t *i2cTestAdaptor) Name() string          { return t.name }
 func (t *i2cTestAdaptor) SetName(n string)      { t.name = n }
 func (t *i2cTestAdaptor) Connect() (err error)  { return }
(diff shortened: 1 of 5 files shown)


deadprogram:
Hi @erkkah thanks for working on this, and it is a great start.

Here are a few items of feedback:

- I think it would be better to extend the current api using functional options to allow passing in a specific bus if desired but use the default bus on the target platform as a default if you pass no params ala @davechaney article https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis

So for https://github.com/erkkah/gobot/blob/tsl2561-driver/drivers/i2c/tsl2561_driver.go#L128 it would be possible to keep the current interface to just pass in the adaptor, along with any extra options for example: 

If we were using the CHIP adaptor

```
a := chip.NewAdaptor()
```

You could have pure defaults:

```
d := NewTSL2561Driver(a)
```

Or we can pass what we need:

```
d := NewTSL2561Driver(a, i2c.TSL2561AddressLow)
```

or

```
d := NewTSL2561Driver(a, i2c.TSL2561AddressLow, i2c.Bus(1))
```

- The file-based implementations are needed for i2c platforms that do not support SMBUS, so perhaps `SMBusOperations` is not the best name, perhaps `I2CBusOperations`? Certainly a good idea to make the convenience functions available.

- I do not think we want a separate i2c API for some devices, and a different one for others. I assume you implemented this in parallel just for exploring the ideas.

- For this PR to land, we would of course need to implement it for all current platforms that support i2c & all i2c drivers.

What do you think?

erkkah:
Hi @deadprogram, thanks for the feedback, and yes, I did this in parallel for experimenting locally, and getting something in place to talk about.

- Functional options should work fine. Thanks for the link! Do you have a suggestion for an interface for the option functions? I'm thinking that the first option in your example, `i2c.TSL2561AddressLow`, would operate on the `TSL2561Driver`, while the function returned by `i2c.Bus(1)` needs to be generic to all different drivers.

- I'm not sure what you mean about the file-based implementation vs the convenience functions. If drivers are supposed to be platform agnostic, they cannot used any of the SMBUS based functions if SMBUS is not supported. Maybe I'm just confused :) Renaming the interface sounds like a good idea though.

- The main problem I have with the `Read(b []byte)`  file based fallback is that the contracts to `file.Read()` and `ReadBlockData()` are different. The first does not use the command byte and the second _might_. That will cause drivers to behave differently on different platforms.

- I'd be happy to take a stab at refactoring existing platforms and devices if we can sort out what we want.

deadprogram:
Oops, just noticed this was awaiting some comment back, sorry @erkkah 

- Regarding functional options, they will need to be scoped to the `i2c` package since that is where all of the drivers "live". I suggest i2c.<Drivername><optionname> e.g. `i2c.TSL2561AddressLow` for device scoped options, and just i2c.<optionname> for options that apply to all i2c drivers.

- I just meant that we want to make sure the register convenience functions such as `WriteRegisterByte()` work with both SMBUS as well as the sysfs file writes, don't we?

- The file based Read/Writes just assume that the first byte is the register, if the target i2c device uses registers. If you look at the i2c specs, "register" just means an agreement to use the first byte that way, it is not a true standard as we might consider it

- Please oh please do take a stab at it, and I can help out however you like. :)

erkkah:
No worries about the delay, @deadprogram, I've been busy :)

- Re functional options, that makes perfect sense. My question was more about the actual interface of the argument functions. The driver generic case might look something like this:
```
func Bus(bus int) func(s *I2cSettings) error {
    return func(s *I2cSettings) {
        s.setBus(bus)
    }
}
```
with the driver specific setting looking like this:
```
func TSL2561AddressLow(s *TSL2561Settings) {
        s.setAddressLow()
}
```
In which case the constructor argument would need to use the more generic type, and we would have to type cast / switch to implement the "apply" - loop. Maybe I'm just slow.
```
func NewTSL2561Driver(a I2c, options ...func(*I2cSettings)) *TSL2561Driver {
}
```
Having looked through different driver implementations, I can see that they use a number of different  schemes for setting options. I think the more generic handling of device driver options should be handled separately from the I2C changes. No need to refactor both at the same time.

- I'm not sure falling back to generic I2C is a good idea, but please tell me if I'm misinformed. Reading https://www.kernel.org/doc/Documentation/i2c/smbus-protocol and https://www.kernel.org/doc/Documentation/i2c/dev-interface, I find these sections helpful in sorting out compatibility and fallback solutions:

> Note that only a subset of the I2C and SMBus protocols can be achieved by
the means of read() and write() calls. In particular, so-called combined
transactions (mixing read and write messages in the same transaction)
aren't supported. For this reason, this interface is almost never used by
user-space programs.

> Some adapters understand only the SMBus (System Management Bus) protocol,
which is a subset from the I2C protocol. Fortunately, many devices use
only the same subset, which makes it possible to put them on an SMBus.

> If you write a driver for some I2C device, please try to use the SMBus
commands if at all possible (if the device uses only that subset of the
I2C protocol). This makes it possible to use the device driver on both
SMBus adapters and I2C adapters (the SMBus command set is automatically
translated to I2C on I2C adapters, but plain I2C commands can not be
handled at all on most pure SMBus adapters).

In my case, I wanted to read a word from a specific register. The device datasheet recommended using SMBus Read Word. I could not get it working using block read or raw read.

That to me indicates what the linux docs suggests, that different devices need different methods of communication, and that it is best to expose those methods to the gobot device drivers. They are not simply convenience methods.

The only SMBus methods that directly map to raw read/write are "Send Byte" and "Receive Byte", but I think we should keep Read and Write, and let them do just that, always talk raw I2C.

erkkah:
Hey, @deadprogram, I'm closing this PR now, but I'll open a new one with a somewhat complete refactoring of all devices and platforms to support more than one I2C bus and a more detailed interface.
No changes to option handling, all devices should work exactly like before.

PR #360, opened 2017-01-24
Adds support for collision detection for Sphero BB8 (& Ollie)

### LINKED ISSUE #349: Sphero BB-8 driver - does it support collision detection?
Not sure if I am missing something obvious, but does BB8Driver support collision detection?

I see that sphero.SpheroDriver has support for collision events. But BB8Driver does not use SpheroDriver at all. It basically uses ollie.Driver which does not seem to handle collisions as well. At least, that is how I am reading the code. Please correct me if I am missing anything.

durgeshm:
Closes https://github.com/hybridgroup/gobot/issues/349

### DIFF
--- a/examples/bb8-collision.go
+++ b/examples/bb8-collision.go
@@ -1,0 +1,35 @@
+package main
+
+import (
+	"fmt"
+	"os"
+
+	"gobot.io/x/gobot"
+	"gobot.io/x/gobot/platforms/ble"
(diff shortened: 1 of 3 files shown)


deadprogram:
Thanks very much for this @durgeshm it is appreciated. Also looks like a great start to implementing more of the response messages for the other sensors.

Merging.

PR #364, opened 2017-02-06
I2C support for several buses and more I2C operations

erkkah:
A new attempt to add support for more than one I2C bus per platform, and provide support for more fine grained operations than Read/Write.

I have not changed any default behavior or added options for selecting other buses than the default. Also, there are several opportunities for more refactoring in the different drivers, to make use of the new I2C operations.

I am unsure if the implementation of block operations on "firmata" makes sense or not. Maybe they should not be supported.

Tests pass on my end, but needless to say, I have not been able to test all platforms and devices.

Grateful for comments!


deadprogram:
Hi @erkkah this is looking pretty amazing. I like the additional operations, perhaps the methods that you named `ReadByteData` etc should be named like `ReadRegisterByte()`, `ReadRegisterWord()`. I think that might be a little more obvious as to their function. What do you think of that?

Great idea to initial leave the current behaviour, and then extend to allow for multiple simultaneous bus usage.

The register operations that you made for Firmata are absolutely needed, and the implementations look OK. I am going to test all of this in more depth this week.

Very exciting progress!

deadprogram:
So far I've tested on Firmata and Edison. Had to fix a couple small things surfaced from these changes, but so far no problems with the new code itself.

Today I will test a few more platforms, I'll post updated status here.

deadprogram:
C.H.I.P. worked fine. For Beaglebone I had to make some corrections to match the proper mappings.

erkkah:
Hi there.
Thanks for the nice feedback.
Great to see that you have time to pick this up now, and that it's mostly working so far!

I'm all for name changes, if they make more sense. I kept names close to the actual protocol operations, to make it easier if someone is sitting with a device datasheet that talks about a specific I2C/SMBus operation.

I agree that the names you suggest better describe what the methods actually do.

Let me know if I can assist on my end.

deadprogram:
You make a good point about sticking to the SMBus-style naming. If we do change those names, which I am not completely convinced of since, we can do in a separate PR.

Now merged into `dev`.

Great job @erkkah thanks for taking this task on! Closing this PR now.

erkkah:
Thank you, @deadprogram, looking forward to the next steps!

deadprogram:
@erkkah you can check out my WIP on the I2cConfig in https://github.com/hybridgroup/gobot/tree/feature/i2c-bus-param branch.

erkkah:
Great, I'll have a look!

PR #369, opened 2017-02-16
Allow NATS options to pass in the NATS adaptor for TLS support.

dnishimura:
This pull request is to allow nats.Option to be passed in to the NATS adaptor to allow TLS support. I've also added some test TLS files for a root CA, server cert/key and client cert/key. The README was a little outdated, so I fixed that as well as added some documentation pertaining to this pull request.

In another future pull request, I'd also like to fix up the unit tests. I read in the comments that ideally the unit tests should run without a server. However, I looked at the NATS Go client repo and they actually run a server in their unit tests, so I was hoping to do the same if that's OK. That way we can truly test the adaptor connection, especially in regards to TLS.

Signed-off-by: Daniel Nishimura <dnishimura@gmail.com>


deadprogram:
Hi @dnishimura first of all thanks for this PR. I'm going to get it into an immediate point release momentarily.

Regarding running an actual NATS server as part or our CI test, I can see where NATS themselves would want to do so for the NATS CI tests, since it is the central point of the NATS project. 

However, for Gobot, for our own CI we only want to test the interfaces for external systems, not the external systems themselves. The correct form for full integration testing from Gobot's POV would probably be running tests right on some boards directly (raspi, edison, etc).

So if someone wanted to create a separate repo with a NATS integration test, that sounds like a way to accomplish such a thing, but launching a NATS instance as part of Gobot's own CI does not seem correct to me.

Anyhow, I hope my reasoning is clear on that. And thanks again for the PR!

dnishimura:
Wow that was the fastest merge ever 😁. Happy to contribute!

Issue #372, opened 2017-02-20
I2C WriteBlockData not working as expected


owenwaller:
Hi,

As discussed in [this](https://groups.google.com/forum/#!topic/golang-nuts/0_cZNZ3vykM) thread on the golang-nuts list.

I am using GoBot v1.2 that has the rewritten i2c subsystem, to try to drive a [PiGlow](https://shop.pimoroni.com/products/piglow) (based on a SN3218 IC) connected to a Raspberry Pi 1.
The code is written using the "metal bot" style as I would like to reuse this as part of another project.

My sample code is available on the Go playground as:
[https://play.golang.org/p/Xma_6zBtT2](https://play.golang.org/p/Xma_6zBtT2)

When I run this code it shows no output. None of the error values are non-nil and led zero does not light on the PiGlow.

As far as I can tell I am communicating with the PiGlow over i2c correctly. My code is based on that of Toon Schoenmakers [code](https://github.com/schoentoon/piglow). This code is not based on the GoBot 12c subsystem. However. Toon's test code works correctly and drives the PiGlow as expected.

Can someone with more experience of GoBots i2c subsystem make a suggestion as to what I might be doing wrong, or if this is a new bug in the i2c subsystem on a RaspPi.

Thanks

Owen

erkkah:
Hi there.

A difference from Toon's code is that you do a "reset" and then return. I don't know how the PiGlow works, but it seems as you init everything, turn on a light, and then reset everything again before exiting. Don't know what "reset" does, but it's not used in Toon's example.

Possibly the default raspi bus in GoBot is not the same as bus 1 used in the other code. But you should have gotten errors then, I guess.

owenwaller:
Hi Erkkah,

Firstly, if I take out the reset behaviour (aka the call to PiGlow.Halt()) I see the same behaviour - i.e. nothing. So it's not the reset call that is the problem. Although you are correct that pointing it in does seem a bit pointless if I want the led to stay on.

Secondly there is only one i2c device on the Rasp Pi 1 and its at /dev/i2c-1.
If I print the result of Adaptor.GetDefaultBus() call as used in the Adaptor.GetConnection call it will return 1.
Toon's code also uses 1 as the i2c bus number.

In fairness, I do believe that on a Rasp Pi 2 and 3 the default i2c bus is in fact 0.

Owen

deadprogram:
Hi, everyone

[...]

PR #378, opened 2017-02-22
I2C use PEC for block transfers if available

erkkah:
Maybe this fixes the issue in #372?


deadprogram:
Looks like the failing check is due to code coverage % going down?

erkkah:
Needs more testing, closing for now.

PR #382, opened 2017-02-28
Hardware PWM support for the C.H.I.P

erkkah:
This PR adds support for PWM0 on the C.H.I.P.

This is one way of doing it, and I have some questions, not sure if the implementation is ok for gobot.

To be able to use PWM0, a device tree overlay is needed in most cases (unless you have a modified kernel). The PR includes an overlay manager that compiles and installs the required overlays.

Now the questions, is this OK?
- The overlay manager depends on the external `dtc` command, as described in the README.md
- The overlay sources files are included as string constants with license info and all
- Using the overlay manager requires root access
- Don't know how to write tests for the overlay or pwm parts, since they operate directly with the file system. Are there examples in gobot where this is done?

### DIFF
--- a/platforms/chip/README.md
+++ b/platforms/chip/README.md
@@ -53,6 +53,63 @@ func main() {
 }
 ```
 
+## Using the overlay manager
+
+The C.H.I.P platform adapter includes an overlay manager for building
+and installing device tree overlays for optional features PWM0 and SPI2.
+
+```go
+package main
+
+import (
+    "fmt"
+    "log"
+    "time"
+
+    "gobot.io/x/gobot"
+    "gobot.io/x/gobot/drivers/gpio"
+    "gobot.io/x/gobot/platforms/chip"
+)
+
+func main() {
+    chipAdaptor := chip.NewAdaptor()
+
+    if err := chip.BuildAndInstallOverlays(); err != nil {
+        log.Println(err)
+        log.Fatal("Failed to build and install overlays")
+    }
+
+    if err := chip.LoadOverlay("PWM0"); err != nil {
+        fmt.Printf("Failed to load overlay: %v\n", err)
+        log.Fatal(err)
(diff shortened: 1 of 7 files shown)


deadprogram:
Hi, @erkkah 

Was just looking this PR over. Seems to me that the device tree overlay should be handled by the developer separately, not done by Gobot itself. You can do an lazy init of the PWM so it is not initialized by default, the assumption being that it is probably not installed. If the developer tried to use `PWM0` then it could try to perform initialization, again the assumption being that the developer has installed the device tree overlay themselves.

If anything, perhaps adding the overlay builder to Gort? In any case, I think that it probably does not belong in Gobot itself.

What do you think?

erkkah:
Hi, @deadprogram.

Thanks for the feedback. I think it makes sense to move the overlay manager out, I just added it because it was a natural place for me to put it since I need it. I'll move it to a separate package, outside of Gobot.

I will give the lazy init suggestion a try, and resubmit.

Don't know enough about Gort to know what that would mean :) I'll start with just moving it out of Gobot.

deadprogram:
Hi @erkkah Gort is our CLI interface for "RobotOps": https://github.com/hybridgroup/gort

Seems like a `gort chip install pwm` command would be ideal here! :)

erkkah:
Hi @deadprogram. I'll close this PR and resubmit from a clean branch instead.

Re Gort, what would be the expected outcome of a `gort chip install pwm` command? Would it simply do the overlay juggling to get the device in place, or would it also modify init scripts to make it permanent?

In other words, would it be expected that a user would have to run the command on each boot?

deadprogram:
Hi @erkkah ideally the command would be run 1 time and then would be "permanent".

I've added an issue to Gort here to discuss: https://github.com/hybridgroup/gort/issues/49

PR #384, opened 2017-03-13
Fix go test ./...

maruel:
Make all examples to not be built by default by adding the build tag 'example'.
Some files were automatically reformatted by goimports upon saving.


deadprogram:
Thanks @maruel looking forward to having Gobot tests run easily using https://github.com/maruel/gohci

PR #392, opened 2017-03-23
Fix neurosky example

b3b:
Fix `go run examples/neurosky.go` error: ``examples/neurosky.go:34: neurosky.EEG is not a type``

neurosky.EEG struct was renamed to neurosky.EEGData : [neurosky_driver.go](https://github.com/hybridgroup/gobot/blob/953c3254e77c4f4a7dd0a03ef7afd25cf463879a/platforms/neurosky/neurosky_driver.go#L67)


deadprogram:
Hi, @b3b thank you for these corrections! Merging.

PR #395, opened 2017-03-25
Significantly speed up travis and make runs

maruel:
- Inline ci/travis.sh into .travis.yml
- Use two go test steps instead of once for speed up
- Drop the whitelist and use all the packages as listed from $(go list ./...).
  This removes the need to do bookeeping in both .travis.yml and Makefile during
  refactoring.
- Trimmed the Makefile significantly. Removed "go test -i" as while this is
  extremely helpful when cross compiling, it doesn't help when compiling for the
  host.
- Added deps as a phony target, made this step faster. I'm not sure of the
  usefulness of this step though and would recommend to remove it.
- 'make examples' is currently broken so I didn't add it to .travis.yml.

I experimented a bit on travis with go test on another similar project, timings
for the go test step only:
- Using both -race and -coverprofile with the loop over $(go list ./...) took
  190s
- Using -coverprofile alone with the loop over $(go list ./...) took 10~11s
- Using -race ./... took 5s

It means that using one after the other takes 16~17s while running with both
flags takes 190s. A 91% speedup. Add my previous commit that cut off the test
runtime by half, the total speed will be appreciable.

Feel free to test then revert if it doesn't give the expected speedup.


maruel:
Oh I forgot the "go get golang.org/x/tools/cmd/cover" in .travis.yml is broken, this line can be removed.
"go test -race ./..." failed: https://travis-ci.org/hybridgroup/gobot/jobs/214951802  :(

maruel:
FTR, coverage took 66s and race took 57s, so 123s total. The build itself took 4m2s due to the significant overhead.
Recent runs:
https://travis-ci.org/hybridgroup/gobot/builds/214412620 took 5m35s
https://travis-ci.org/hybridgroup/gobot/builds/214220640 took 5m6s
https://travis-ci.org/hybridgroup/gobot/builds/214219508 took 4m53s

so the saving is maybe around 20% after all, not as much as I'd expect but still better than nothing.

deadprogram:
I did not merge this yet, for the main reason that the race conditions that are identified by the new test commands should probably be fixed first, don't you think?

maruel:
Yes. :) I haven't taken time to investigate the race conditions now exposed.

I wonder why you are not seeing these otherwise: I suspect that using both -coverprofile and -race reduces the effectiveness of race detection but I haven't tried to confirm.

deadprogram:
FInally got all of the race conditions and resolved, and have merged this. Thanks again @maruel for your fine efforts!

maruel:
Humm I'm seeing [3m30s for the commit merge](https://travis-ci.org/hybridgroup/gobot/builds/217886322) but [2m17s for the run before](https://travis-ci.org/hybridgroup/gobot/builds/217883388).

The later are much faster than what I had previously anecdotally observed (which was in the 5minutes range) so maybe other things changed in the meantime(?) My patch increases the test coverage a bit but I wouldn't expect this much slow down.

Anyhow I think it's good to settle for coverage then race as it seems to have exposed more race conditions than before, a few whitelisted packages had race conditions but these weren't detected but I'm concerned about the overall slowdown.

deadprogram:
Hmmm, I see what you mean with the slowdown. How do you suggest we proceed?

maruel:
https://travis-ci.org/hybridgroup/gobot/jobs/217883390
`./ci/travis.sh` took 36s for 37 packages

https://travis-ci.org/hybridgroup/gobot/jobs/217886327

- `gofmt` check took 0.5s so negligible
- `go test -cover` took 62s for 43 packages
  - if you look at each test runtime, it doesn't add up. It seems to have hidden overhead.
- `go test -race` took 58s
  - you can clearly see that the runtime on each package is much longer than it was with `./ci/travis.sh`, which means that travis.sh wasn't fully testing for race conditions, as exposed with the fixes you did.
  - the sum of the displayed runtime make sense with the overall runtime so there doesn't seem to be much overhead.

My current hypothesis is that maybe I was wrong on travis and that a `go test -i` may be helpful when using `go test -cover` but likely has no effect when using `go test -race`. Still the gain from the race condition detection to actually work overweight the slow down. So my recommendation would be to try to add `go test -i ./...` back as a step [before line 26](https://github.com/hybridgroup/gobot/blob/dev/.travis.yml#L26) and see if it improves performance of the cover step..

deadprogram:
Hi @maruel so I tried your suggestion, and it had no effect on the run times for the coverstep:

https://travis-ci.org/hybridgroup/gobot/jobs/217886327 without -i flag

https://travis-ci.org/hybridgroup/gobot/jobs/218117124 with `go test -i ./...`

maruel:
I think the current state is the best solution as it ensures that everything is compilable and is well race tested but that's really up to you. It is clear to me that using simultaneously -cover and -race did not detect race conditions so it's still preferable to run the tests twice.

deadprogram:
Yes, I quite agree that the fact that any race conditions are now exposed by the current tests is much preferable. So at least for now, we're good. Thanks again for your help!

PR #396, opened 2017-03-27
Add crude travis check for gofmt; format all sources

maruel:
The check ensures no gofmt regression occurs. Formatting the code makes everyone happy.

Simply ran the command: `gofmt -s -w .`


deadprogram:
Thanks for this, @maruel now merging.

Issue #374, opened 2017-02-20
Firmata fails to start a connection with Arduino mega2560


matipan:
Hi!
I'm trying to run the [blinking example](https://gobot.io/documentation/platforms/arduino/) with an Arduino MEGA2560.
These are the steps I followed:
```
$ gort scan serial
/dev/cu.Bluetooth-Incoming-Port		/dev/tty.Bluetooth-Incoming-Port
/dev/cu.usbmodem1421			/dev/tty.usbmodem1421
```
So I changed the NewAdaptor to `firmata.NewAdaptor("/dev/tty.usbmodem1421")`.
Then `gort arduino install` and it installed `avrdude-6.3` successfully.
Now I run:
```
$ gort arduino upload firmata /dev/tty.usbmodem1421 -b mega2560
avrdude: AVR device initialized and ready to accept instructions

Reading | ################################################## | 100% 0.01s

avrdude: Device signature = 0x1e9801 (probably m2560)
avrdude: reading input file "/var/folders/2l/2x1l6t5530s25fzgvyfzdp_r0000gn/T/816739554"
avrdude: writing flash (12224 bytes):

Writing | ################################################## | 100% 1.97s

avrdude: 12224 bytes of flash written
avrdude: verifying flash memory against /var/folders/2l/2x1l6t5530s25fzgvyfzdp_r0000gn/T/816739554:
avrdude: load data flash data from input file /var/folders/2l/2x1l6t5530s25fzgvyfzdp_r0000gn/T/816739554:
avrdude: input file /var/folders/2l/2x1l6t5530s25fzgvyfzdp_r0000gn/T/816739554 contains 12224 bytes
avrdude: reading on-chip flash data:

Reading | ################################################## | 100% 1.57s

avrdude: verifying ...
avrdude: 12224 bytes of flash verified

avrdude done.  Thank you.
```
According to the output of that command, the firmata software was successfully loaded. So the only thing left to do is run the program:
```
$ go run main.go
2017/02/20 20:10:49 Initializing connections...
2017/02/20 20:10:49 Initializing connection Firmata-3BD25BC8D8022818 ...
2017/02/20 20:10:49 Initializing devices...
2017/02/20 20:10:49 Initializing device LED-4A0D0593775CB601 ...
2017/02/20 20:10:49 Robot bot initialized.
2017/02/20 20:10:49 Starting Robot bot ...
2017/02/20 20:10:49 Starting connections...
[...]

Issue #376, opened 2017-02-22
Gobot hangs while finalizing connections with Firmata on OSX


matipan:
Hi there!
This happened to me wihle running [a simple go program](https://gist.github.com/matipan/af6f0c5a4fa826984267e070d62027bd) :
```
2017/02/22 11:01:13 Initializing connections...
2017/02/22 11:01:13 Initializing connection Firmata-72720CD878A08339 ...
2017/02/22 11:01:13 Initializing devices...
2017/02/22 11:01:13 Initializing device LED-3FA3E55721925098 ...
2017/02/22 11:01:13 Robot bot initialized.
2017/02/22 11:01:13 Starting Robot bot ...
2017/02/22 11:01:13 Starting connections...
2017/02/22 11:01:13 Starting connection Firmata-72720CD878A08339 on port /dev/tty.usbmodem1411...
2017/02/22 11:01:18 Starting devices...
2017/02/22 11:01:18 Starting device LED-3FA3E55721925098 on pin 13...
2017/02/22 11:01:18 Starting work...

2017/02/22 11:01:21 HELLO

2017/02/22 11:01:22 HELLO
^C2017/02/22 11:01:24 Stopping Robot bot ...

2017/02/22 11:01:25 HELLO

2017/02/22 11:01:25 HELLO


2017/02/22 11:01:25 HELLO
2017/02/22 11:01:36 HELLO
2017/02/22 11:01:36 HELLO
2017/02/22 11:01:37 HELLO
2017/02/22 11:01:38 HELLO
2017/02/22 11:01:39 HELLO
^C^C^C^C
2017/02/22 11:01:45 HELLO
```
The go program never stops(even when I repeatedly send the interruption signal), It keeps executing what is inside the work func.

I think the problem is inside the [Stop](https://github.com/hybridgroup/gobot/blob/master/robot.go#L196) function, when it tries to finalize all the connections, for some reason it never returns from that method.
Here is the stack trace:
```
panic: STACK TRACE

goroutine 1 [running]:
github.com/matipan/gobot/vendor/gobot.io/x/gobot.(*Robot).Stop(0xc4200b0300, 0xc4200b0540, 0x0)
	/Users/Matias/Code/GoWorkspace/src/github.com/matipan/gobot/vendor/gobot.io/x/gobot/robot.go:196 +0x19e
[...]

PR #403, opened 2017-04-07
core: use 10-buffered chans for events, see #374

rkjdid:
make eventer.in & outs buffered channels.

this fixes some race conditions where client.Connect() would hang waiting on locks & unbuffered chans

see issue #374

### DIFF
--- a/eventer.go
+++ b/eventer.go
@@ -18,6 +18,8 @@ type eventer struct {
 	eventsMutex sync.Mutex
 }
 
+const eventChanBufferSize = 10
+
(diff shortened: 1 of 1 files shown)


deadprogram:
Thank you, @rkjdid for this. I think it might also address https://github.com/hybridgroup/gobot/issues/376

Now merging.

rkjdid:
not certain it does, I'm having the issue you mention also when interrupting.
I think it doesn't happen if explicitly calling robot.Stop()

rkjdid:
nvm, it might be the case. I had a wrong use-case

PR #408, opened 2017-04-16
mqtt: add a (topic, payload) event type.

nzmichaelh:
Used for wildcard subscriptions such as 'device/topic/#'.  In particular, I'm using it for 'set' support on parameters so something exported as device/topic/param can be set via device/topic/param/set.

Signed-off-by: Michael Hope <mlhx@google.com>


deadprogram:
Hi, @nzmichaelh thank you for working on this.

I am just wondering if perhaps this should be the default behavior to always just return the topic along with the data?

Also, seems like this same functionality is also needed in the Driver, however it gets implemented: https://github.com/hybridgroup/gobot/blob/dev/platforms/mqtt/mqtt_driver.go#L77

What do you think?

nzmichaelh:
Hi @deadprogram thanks for the response!  I'm happy to change On(payload) to be On(topic, payload).  Its a breaking change to the API, but if you're happy with that then I'm happy with that.

deadprogram:
I do not like breaking the API, but wildcards are an essential capability of MQTT. There does not seem to be any way around this, other than your current PR, or breaking the API. I see this as an omission on my part when implementing the original API.

Other messaging systems also have the same wildcard feature. We should also change the NATS adaptor https://github.com/hybridgroup/gobot/blob/dev/platforms/nats/nats_adaptor.go#L115 to do the same as whatever we decide to do here.

nzmichaelh:
Done, PTAL.  The NATS patch is in #409.

deadprogram:
OK @nzmichaelh you will probably want to kill me, but I have yet another idea :)

What if we did this?

```go
mqttAdaptor.On("hello", func(msg mqtt.Message) {
   fmt.Println(msg.Topic(), msg.Payload())
})
```

`mqtt.Message` in this case means https://github.com/eclipse/paho.mqtt.golang/blob/master/message.go#L24

It allows your use case, plus a few other we do not yet know about for speccific purposes. Basically just  exposes everything.

deadprogram:
We could probably declare `type Message paho.Message` in the Gobot `mqtt` package, to allow Gobot devs to refer to `mqtt.Message`.

deadprogram:
Hello @nzmichaelh what did you think of my idea?

nzmichaelh:
Hey @deadprogram.  Yip, LGTM.  Working on it now.

nzmichaelh:
Done.  PTAL.  I'll update the NATS branch after this is merged.

deadprogram:
Thanks @nzmichaelh this looks good. Now merging.

deadprogram:
FYI @nzmichaelh I updated the examples that use mqtt to match the new signature here: https://github.com/hybridgroup/gobot/commit/311bfcfb451fb423f70dcb002b8355066ec25859

nzmichaelh:
Thanks!  I spotted the README but not the examples :)

Issue #287, opened 2016-06-11
why is raspberry pi PWM limited between 0 and 255?


nhooyr:
The underlying pi-blaster daemon supports 1000 steps. Why only limit PWM write values between 0 and 255?

deadprogram:
Hi, @nhooyr the interface only supports that many, to provide compatibility with other interfaces such as Firmata etc. So really no good reason at all.

Perhaps a "PWMScale" method or something like that to allow setting different variations of PWM control. What do you think? 

I'd be happy to accept a PR for the Raspi platform which added this. ;)

deadprogram:
I'm going to close this issue due to inactivity, please reopen if needed. Thanks!

verrol:
Hi,

Please re-open this issue to add some kind of scale method for PWM modules that uses more than 8bits.

Since gpio.DirectPinDriver.PwmWrite() uses a byte value, the max duty cycle on an ESP8266 is ~25%. 

The ESP8266 appears to have a 10-bit resolution, so when specifying 255, the duty cycle is only a fraction of the max. Please see the attached screenshot from my logic analyzer.

The following code was used to test all 4 PWM pings for the ESP8266.

Thanks,

.v

```go
package main

import (
	"flag"
	"log"
	"time"

	"gobot.io/x/gobot"
	"gobot.io/x/gobot/drivers/gpio"
	"gobot.io/x/gobot/platforms/firmata"
)

/*

 NodeMCU Pin  | GPIO   | Function Pin
--------------+--------+-------
 10           | 12     | PWM0
 13           | 15     | PWM1
 9            | 14     | PWM2
 16           | 4      | PWM3

*/
const (
	pwm0Pin     = "12"
	pwm1Pin     = "15"
	pwm2Pin     = "14"
	pwm3Pin     = "4"
	defaultPort = "10.10.100.175:3030"
)

func main() {
	flag.Parse()

	port := defaultPort

	if len(flag.Args()) == 1 {
		port = flag.Args()[0]
	}

	log.Printf("Using port %v\n", port)

	board1 := firmata.NewTCPAdaptor(port)
	pwm0 := gpio.NewDirectPinDriver(board1, pwm0Pin)
	pwm1 := gpio.NewDirectPinDriver(board1, pwm1Pin)
	pwm2 := gpio.NewDirectPinDriver(board1, pwm2Pin)
	pwm3 := gpio.NewDirectPinDriver(board1, pwm3Pin)

	var speed byte

	work := func() {
[...]

PR #410, opened 2017-04-19
Feature/pwm period

bezineb5:
Hello,

I've been trying to use servos with Gobot on the Raspberry Pi, and I think it's currently a bit too limited to support all cases. I'm proposing this change to offer better support. The improvements are:
* Support for custom angle range (you can say that you servo can go from -140 to +150 degrees)
* Support for custom PWM timings, as servos are not using the same periods for the signal.
* Added interface to set the PWM signal period in seconds which removes the need for a ServoWriter interface. It means that any platform which supports PWM supports servos, with less code.
* That change also improves the precision of the positioning, as described in issue #287 

The last point is a breaking change, and currently only works on Raspberry Pi - I don't know the other platforms. You have to consider that this is a work in progress.

The creation of a new instance would be like that:
`servo := gpio.NewAngularServoDriver(r, "40", gpio.WithServoAngles(-80.0, 80.0), gpio.WithPeriodRange(0.7/1000.0, 2.3/1000.0))`
With the gpio.WithServoAngles and gpio.WithPeriodRange being optional, defaulting to current values in the servo implementation.

And the interface to move the servo remains unchanged (except that angles are float):
`servo.Move(45.8)`

What do you think? Do you think it could be a good change for Gobot?

### DIFF
--- a/drivers/gpio/angular_servo_driver.go
+++ b/drivers/gpio/angular_servo_driver.go
@@ -1,0 +1,94 @@
+package gpio
+
+import "gobot.io/x/gobot"
+
+// AngularServoDriver Represents a Servo
+type AngularServoDriver struct {
+	name       string
+	pin        string
+	connection PwmWriter
+	gobot.Commander
+	CurrentAngle float64
+	MaxAngle     float64
+	minPeriod    float64
+	maxPeriod    float64
+}
+
+// NewAngularServoDriver returns a new ServoDriver given a ServoWriter and pin.
+//
+// Adds the following API Commands:
+// 	"Move" - See ServoDriver.Move
+//	"Min" - See ServoDriver.Min
+//	"Center" - See ServoDriver.Center
+//	"Max" - See ServoDriver.Max
+func NewAngularServoDriver(a PwmWriter, pin string, maxAngle float64, minPeriod float64, maxPeriod float64) *AngularServoDriver {
+	s := &AngularServoDriver{
+		name:         gobot.DefaultName("AngularServo"),
+		connection:   a,
+		pin:          pin,
+		Commander:    gobot.NewCommander(),
+		CurrentAngle: 0,
+		MaxAngle:     maxAngle,
+		minPeriod:    minPeriod,
+		maxPeriod:    maxPeriod,
+	}
+
+	s.AddCommand("Move", func(params map[string]interface{}) interface{} {
+		angle := params["angle"].(float64)
+		return s.Move(angle)
+	})
+	s.AddCommand("Min", func(params map[string]interface{}) interface{} {
+		return s.Min()
+	})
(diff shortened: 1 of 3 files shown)


deadprogram:
Hi @bezineb5 thanks for your thoughts and work on this.

I am in the process of extracting the PWMPin implementation and interface into a shared part of the `sysfs` package. Please take a look at https://github.com/hybridgroup/gobot/blob/dev/sysfs/pwm_pin.go

I think we can accomodate some of what you are proposing in a way that does not require an entirely new driver, but instead some of that work can be incorporated into the existing Servo driver with some additional options.

As far the as ServoWriter interface, there are platforms such as Firmata that perform specific functionality in the microcontroller specific to servo control. Seems like some additional options could solve this problem in the general case, while allowing for higher resolution servos.

What do you think?

bezineb5:
Hi @deadprogram,
Thanks for the reply!
Yes, for sure, the AngularServoDriver should be integrated in the regular Servo driver, it's just that I didn't want to break everything. By the way, shouldn't gobot also support continuous servos?
I had a look at your new PWMPin implementation, and I fully agree that the new period/duty cycle methods allow the finer control I'm looking for. The only think I don't understand is the use of a string for the period. Why not use a float64/float32, in seconds or milliseconds?

deadprogram:
Hi @bezineb5 agreed about not using string for `Period`. But I was thinking more of switching to uint32 as the Linux expects that param to be an integer expressing the period in nanoseconds, same as duty cycle.

In any case, I will work a little more on the PWMPin implementation to make that change. Anything else you'd like to see in the implementation?

deadprogram:
@bezineb5 I've made those changes to the `dev` branch. Please take a look at the latest implementation.

deadprogram:
Hi @bezineb5 not sure if you had a chance to look over the updated PWMPin implementation, but I think it is completed, at least for now.

I am going to close this PR, please open a new one when you continue working on your servo improvements.

Thanks again for helping!

deadprogram:
There is one last round of refactoing I am doing, specifically to expose the PWMPin pins to programmers via  `PWMPin()` method on the `Adaptor`. My incomplete work is in the https://github.com/hybridgroup/gobot/tree/feature/pwmpin branch.

deadprogram:
Hi @bezineb5 I've completed the refactoring enough that it was merged into the `dev` branch. I've added implementations of the new PWMPinnerProvider & PWMPinner interfaces into all but 1 of the Adaptor's that can support such functionality.

You can use it for your more advanced servo code and other such needs, by obtaining a `PWMPin` from the adaptor like this:

```
a := raspi.NewAdaptor()
pin := a.PWMPin("17")

// get period
period := pin.Period()

// set to 50% duty cycle
duty := period / 2
pin.SetDutyCycle(duty)
```

I think this should accomodate your needs. If you need anything specific from these changes, or just want to comment, please open an issue.

PR #412, opened 2017-04-25
Ads1x15 driver

bezineb5:
Hello,

This commit adds supports for the ADS1015 and ADS1115. All the readings are converted in volts for easy interpretation/consumption by the application.
Please note that it doesn't implement the AnalogReader as it deals with integers, and I didn't know what to return.
I believe that the AnalogReader interface should be modified to use floats and use the volts as it's the SI unit for electric potential.
What do you think?

I've not yet done the unit tests, so it is still work in progress.


bezineb5:
I've just seen that there is an implementation of ADS1015 in dev. Maybe it would be worth merging the 2 versions. My version has the following advantages:
* support of ADS1015 and ADS1115
* conversion of the integer values to volts.

deadprogram:
Hi @bezineb5 thanks for working on this. Second time we are working on the same thing at the same time, we need to coordinate efforts better.

Anyhow, I like several things about your implmentation. Support for both ADS1x15 uints is one. Better validation of params such as data rates is another.

The conversion to V is cool but slightly less interesting, only because my priority was obviously to implement `AnalogReader` interface, which the `dev` branch has. Also, tests.

So I really like a lot of your implementation. How do you suggest we proceed? One option that comes to mind is just to replace the `dev` implementation with your branch, after adding the `AnalogReader` interface and test coverage. Another option would be to incorporate some of your work into the `dev` branch. Not sure that is any easier, and probably harder.

Anyhow, if you are up for it, I'd love to see your branch get the above changes, and then replace the current `dev` implementation. What do you think?

bezineb5:
Actually, it's my fault, I didn't have a look at the dev branch before... Sorry for that! FYI, I have other work in progress for the MPU9250 Gyro/accelerometer/compass and the VL6180X "time-of-flight" chip.
Concerning the use of volts, I think this is useful. Let me try to make my point:
* For example, I'm using an ADS1015 to read the voltage from a power supply (https://www.adafruit.com/product/1944), so that I can detect when the battery is about to run low. In that case, it's meaningful to have volts to interpret the values.
* Using volts for all ADCs make them interchangeable. I have ADS1015 and ADS1115, and I can just switch them with almost no code change (apart from the constructor of the driver).
* During my searches, I came across that article from Adafruit which explains why they decided to use SI units in their drivers. I think that would be a great approach for Gobot, too! https://learn.adafruit.com/using-the-adafruit-unified-sensor-driver/introduction

I like many things in your implementation:
* The nice const names with comments
* The option "WithADS1015Gain"
* The AnalogReader support
* For the difference, I don't really like to use "1" to mean "0-3"... I like what you did with the pin name to make it clearer.

To proceed, do you want me to do the merge? I can do that in the next couple of days and propose you a draft.

Thanks!

deadprogram:
Sounds like you are on the right track to me.

I will hold off on any further changes to ADS1015 pending your PR to reify both versions into a single code base.

Also, looking forward to the MPU9250 & VL6180X implementations!

bezineb5:
Mmmh, sounds like the rebase did not go so well, there are other files in that PR.

drivers/i2c/ads1x15_driver.go
```diff
@@ -0,0 +1,359 @@
...
+
+	// Second case: read the voltage at a specific pin, compared to the ground
+	channel, err = strconv.Atoi(pin)
+	if err != nil {
+		return
+	}
+
+	read, err = d.ReadWithDefaults(channel)
+	if err == nil {
+		value = int(1000.0 * read)
```
deadprogram: I think this line should be:

```
value = int(gobot.ToScale(gobot.FromScale(read, 0, d.gainVoltage[d.DefaultGain]), 0, 1023))
```

drivers/i2c/ads1x15_driver.go
```diff
@@ -0,0 +1,359 @@
...
+	}
+
+	voltageMultiplier, ok := d.gainVoltage[gain]
+
+	if !ok {
+		err = errors.New("Gain must be one of: 2/3, 1, 2, 4, 8, 16")
+		return
+	}
+
+	value = d.converter(data) * voltageMultiplier
```
deadprogram: I think this line should be:

```
value = d.converter(data) / voltageMultiplier * voltageMultiplier
```

deadprogram:
Hi @bezineb5 the tests are failing in Golang 1.7.x due to a difference in the error returned by the parser. Still an error, so probably changing the test would be best.

Also, still need to correct the scaling for the voltage function. When I run the example andhttps://github.com/bezineb5/gobot/blob/ads1x15_driver/examples/raspi_ads1015.go#L22 with the correct `v, _ := ads1015.ReadWithDefaults(0)` line, I get this result:

```
A0 32.752
```

That is what is being returned by the voltage function, since you cannot simply multiply here and obtain the correct result: https://github.com/bezineb5/gobot/blob/ads1x15_driver/drivers/i2c/ads1x15_driver.go#L370

The proper formula for the calculation there should be `value = d.converter(data) / voltageMultiplier * voltageMultiplier`

bezineb5:
Ok, I see your problem: you're testing with fake values which return more than 12 bits of actual precision, right? The ADS1015 is a 12-bit ADC, I'm going to add some ore checks there. It's not a problem of voltageMultiplier.

deadprogram:
No I am using my actual ADS1015 on my Raspberry pi for that.

deadprogram:
Hi @bezineb5 to elaborate a little:

I have a Raspberry Pi 3 with an ADS1015 attached. I am using an analog rotary sensor connected to pin 0 on the ads1015.

This gist shows my output when I run the sample, and turn the knob from one extreme to another: https://gist.github.com/deadprogram/a0e3c649d0fe088b0793a29a400e7bc4

Hopefully that makes clear what is going on here. Thanks again for your help on this!

deadprogram:
Hi @bezineb5 anything more I can do to help this PR land? :)

bezineb5:
Ok, I found it. It was a reading bug for the ADS1015. I managed to get a 1015 and try with both components. That should be good now!

deadprogram:
Tested, and the V is what I expected for my setup. Now merging, thanks again for working on this @bezineb5

PR #415, opened 2017-05-05
Fixed raspi adaptor race conditions

mighe:
Same as PR #414 , but on `dev` branch

### DIFF
--- a/platforms/raspi/raspi_adaptor.go
+++ b/platforms/raspi/raspi_adaptor.go
@@ -122,7 +122,6 @@ func (r *Adaptor) DigitalPin(pin string, dir string) (sysfsPin sysfs.DigitalPinn
 		return
 	}
 
-
 	if err = currentPin.Direction(dir); err != nil {
 		return
 	}
--- a/platforms/raspi/raspi_adaptor.go
+++ b/platforms/raspi/raspi_adaptor.go
@@ -11,6 +11,7 @@ import (
 	"gobot.io/x/gobot"
 	"gobot.io/x/gobot/drivers/i2c"
 	"gobot.io/x/gobot/sysfs"
+	"sync"
 )
 
 var readFile = func() ([]byte, error) {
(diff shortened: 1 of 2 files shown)


deadprogram:
Hi @mighe thanks for resubmitting.

Could you please run gofmt on the files, and then commit those changes? That part of the Travis test suite is failing: https://travis-ci.org/hybridgroup/gobot/jobs/229094248#L898

Thank you.

mighe:
Sorry, I forgot to check "go fmt" in my IDE :no_good_man:

platforms/raspi/raspi_adaptor.go
```diff
@@ -100,18 +116,31 @@ func (r *Adaptor) DigitalPin(pin string, dir string) (sysfsPin sysfs.DigitalPinn
...
-	if err = r.digitalPins[i].Direction(dir); err != nil {
+	if err = currentPin.Direction(dir); err != nil {
 		return
 	}
 
-	return r.digitalPins[i], nil
+	return currentPin, nil
+}
+
+func (r *Adaptor) getExportedDigitalPin(translatedPin int, dir string) (sysfsPin sysfs.DigitalPinner, err error) {
```
deadprogram: Isn't this doing the same thing as the `DigitalPin` method? Seems like should be using that method. What do you think?

mighe: I extracted this part to minimize the mutex locking and make it easier to implement.
Doing so in `DigitalPin` would require either to lock `currentPin.Direction` or unlock the mutex without defer with a more error prone code.
This way `DigitalPin` just call `getExportedDigitalPin` without using the mutex in an explicit way.

If you prefer I can inline the new function, but I think this way is better

deadprogram: You should leave it, your implementation seems quite reasonable.

Seems like this same locking is needed for `PWMPin()`, do you agree?

mighe: in `PWMPin()` I preferred an explicit locking strategy (lines 199 and 200) without creating a new func because in this case we're just creating a new struct and returning it immediately.

deadprogram:
Thanks @mighe now merging.

PR #417, opened 2017-05-06
Use sensor data in tests.

pauleyj:


deadprogram:
Thanks, @pauleyj now merging.

Issue #424, opened 2017-05-29
Humidity is disabled in the BME280 I2C driver because of a missing write operation


ssorc:
Calling the `Humidity()` function on the `BME280Driver` I2C driver results in the error: "Humidity disabled".

This is because the `initHumidity()` function is incomplete; it writes out the humidity data acquisition options, but doesn't then write to the "ctrl_meas" register, as per Section 5.4.3 of the BME280 data sheet:

> The “ctrl_hum” register sets the humidity data acquisition options of the device. Changes to
> this register only become effective after a write operation to “ctrl_meas”.

I have a patch so can generate a pull request if someone else can confirm that this is the case! I have tested this on a Raspberry Pi Model B using the Sparkfun BME280 Atmospheric Sensor Breakout with gobot 1.5.0 and golang 1.7.5.

deadprogram:
Hi @ssorc thanks for the report. I will setup to test this issue shortly and get back to you.

deadprogram:
Hi @ssorc yes, I can confirm that error is being returned.

Interestingly, restarting the program causes the readings to come thru correctly on subsequent runs.

In any case, your PR to address this would be appreciated. Thank you!

ssorc:
The restart may work because when the call to `d.initialization()` in `Start()` is writing to the `ctrl_meas` register (see the `bmp280RegisterControl` register in the BPM280 driver's `initialization()` function) and this may be enough to finally enable the humidity sensor.

I will send through a PR: it's only a couple of lines.

ssorc:
I've submitted the PR.

deadprogram:
Version 1.6.0 is now out, and includes this fix. Now closing, thanks again @ssorc for this!

PR #425, opened 2017-05-31
Issue #424: Enables humidity readings in the BME280 driver by enforci…

ssorc:
…ng the write to the 'ctrl_meas' register, as per Section 5.4.3 of the BME280 data sheet

Signed-off-by: Graeme Cross <graeme@ceriumdesigns.com>

### DIFF
--- a/drivers/i2c/bme280_driver.go
+++ b/drivers/i2c/bme280_driver.go
@@ -7,6 +7,7 @@ import (
 )
 
 const bme280RegisterControlHumidity = 0xF2
+const bme280RegisterControl = 0xf4
 const bme280RegisterHumidityMSB = 0xFD
 const bme280RegisterCalibDigH1 = 0xa1
 const bme280RegisterCalibDigH2LSB = 0xe1
@@ -113,6 +114,7 @@ func (d *BME280Driver) initHumidity() (err error) {
 	d.hc.h5 = 0 + (int16(addrE6) << 4) | (int16(addrE5) >> 4)
 
 	d.connection.WriteByteData(bme280RegisterControlHumidity, 0x3F)
+	d.connection.WriteByteData(bmp280RegisterControl, 0x3F)
 
 	return nil
 }
--- a/drivers/i2c/bme280_driver.go
+++ b/drivers/i2c/bme280_driver.go
@@ -117,9 +117,10 @@ func (d *BME280Driver) initHumidity() (err error) {
 	// The 'ctrl_hum' register sets the humidity data acquisition options of
 	// the device. Changes to this register only become effective after a write
 	// operation to 'ctrl_meas'. Read the current value in, then write it back
-	cmr, err := d.connection.ReadByteData(bmp280RegisterControl)
-	if err != nil {
-		d.connection.WriteByteData(bmp280RegisterControl, cmr)
+	var cmr uint8
+	cmr, err = d.connection.ReadByteData(bmp280RegisterControl)
+	if err == nil {
+		err = d.connection.WriteByteData(bmp280RegisterControl, cmr)
 	}
 	return err
 }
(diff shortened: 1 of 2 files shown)


drivers/i2c/bme280_driver.go
```diff
@@ -113,6 +114,7 @@ func (d *BME280Driver) initHumidity() (err error) {
 	d.hc.h5 = 0 + (int16(addrE6) << 4) | (int16(addrE5) >> 4)
 
 	d.connection.WriteByteData(bme280RegisterControlHumidity, 0x3F)
+	d.connection.WriteByteData(bmp280RegisterControl, 0x3F)
```
deadprogram: Same question here are the above line.

ssorc: I had taken line 117 from the initialisation code that calls into the BMP280 driver's initialisation code. Re-reviewing the BMP280 and BME280 data sheets, a value of `0x3F` does *not* seem correct for either of the lines in question.

Line 117 above would be better replaced with this:
```
// The 'ctrl_hum' register sets the humidity data acquisition options of
// the device. Changes to this register only become effective after a write
// operation to 'ctrl_meas'. Read the current value in, then write it back
cmr, err := d.connection.ReadByteData(bmp280RegisterControl)
if err != nil {
	d.connection.WriteByteData(bmp280RegisterControl, cmr)
}
```

This is the approach taken by other BME280 drivers, including the official Bosch BME280 driver (https://github.com/BoschSensortec/BME280_driver/blob/master/bme280_support.c, lines 172 - 180) as it does not override existing oversampling rates that have been set up for the `ctrl_meas` register.

For the byte written to `bme280RegisterControlHumidity`, the value is the oversampling rate, as defined in Table 20 of the Bosch BME280 data sheet where basically three bits control an oversampling rate of x0 (skipped) to x16.

I would suggest a default oversampling rate of 1 (as per Section 3.5 of the data sheet: "Recommended modes of operation"). 

If you are happy with this approach, I will update the code.

deadprogram: Sounds good to me. Yes, please.

deadprogram: Hi @ssorc I'd really like to merge this in preparation for the next Gobot release. Any chance of getting that update soon? :)

Thank you!

ssorc: Hi @deadprogram; yes; expect an update later today.

drivers/i2c/bme280_driver.go
```diff
@@ -113,6 +114,7 @@ func (d *BME280Driver) initHumidity() (err error) {
 	d.hc.h5 = 0 + (int16(addrE6) << 4) | (int16(addrE5) >> 4)
 
 	d.connection.WriteByteData(bme280RegisterControlHumidity, 0x3F)
```
ssorc: At some point, we should revisit the 0x3F magic constant here:

    d.connection.WriteByteData(bme280RegisterControlHumidity, 0x3F)

which existed before my PR as well as the similar line in the bmp280_driver.go `initialization()` function:

    d.connection.WriteByteData(bmp280RegisterControl, 0x3F)

Being able to specify the oversampling rates for temperature and pressure in the BMP280 driver, as well as humidity in the BME280 driver, would eliminate these magic constants and also ensure that we are using the correct values.

ssorc:
Ignore that last commit; there is an incorrect error check in `initHumidity()` that causes the unit tests to fail. Fixing now.

ssorc:
Updated PR. 
The unit test that was failing now passes but I would welcome feedback if I haven't appropriately modified the `TestBME280DriverStart()` unit test. The change I made is to ensure the `ReadByteData()` call in `initHumidity()` doesn't fail.

deadprogram:
Thanks! Now merging.

PR #426, opened 2017-06-03
Fix for broken PWM on C.H.I.P

erkkah:
This PR fixes the PWM implementation after the recent refactoring:
  * Changed init order to set polarity in "disabled" state
  * Fixed `ServoWrite` to handle the change from frequency to duration


erkkah:
Oops. Had obviously not tested that one enough, also not sure if the same condition as on C.H.I.P holds on other platforms, that polarity cannot be set when PWM is enabled.

erkkah:
Just checked https://www.kernel.org/doc/Documentation/pwm.txt, and the contract holds, cannot set polarity when enabled.

deadprogram:
Thanks @erkkah for the PR! Now merging.

PR #427, opened 2017-06-06
fix signed/unsigned bug

pauleyj:
getBusVoltageRaw and getShuntVoltageRaw should return signed 16-bit integers.


deadprogram:
Thanks @pauleyj now merging.

PR #431, opened 2017-06-20
[Docs] Call event handler from drivers, not core

kpashka:
https://github.com/hybridgroup/gobot-site/pull/59


deadprogram:
Thanks @kpashka

PR #433, opened 2017-06-27
Add the installation steps for the CLI snap

elopio:


elopio:
With the installation steps in the README we will get more testers to install it, so we can later move it to stable.

@deadprogram we also have a new service that will take care of the continuous delivery of the snap every time you make a change in master. You will only have to click a couple of buttons in https://build.snapcraft.io/ Please take a look and let me know what do you think.

deadprogram:
Good idea, @elopio thanks and I'll look into the auto publish.

Can you please make this PR against the `dev` branch to avoid merge conflicts with other upcoming changes?

Thank you!

elopio:
ahh, sorry. Luckily, github makes that very simple to change. Now it's proposed in dev.

deadprogram:
Thanks @elopio now merging.

PR #437, opened 2017-07-16
Add Linux specific step to Intel Curie docs

ljfranklin:
- My Ubuntu 16.04 installation was initially unable to communicate with
  the intel TinyTile with "unable to connect. Perhaps you need to flash
  your Arduino with Firmata?"
- After digging through some forums, the fix is to add some `udev`
  rules. Although I'm not sure of the specific effect of these rules, I can
  confirm the running `sudo go run main.go` is not sufficient to connect to the board.
- After running the included script, unplugging the board and plugging back in, the board connected as expected.
- Rule script:
  https://raw.githubusercontent.com/01org/corelibs-arduino101/master/scripts/create_dfu_udev_rule


deadprogram:
Hi @ljfranklin thanks for this. I installed that udev rule so long ago I had forgotten it.

Now merging. Thanks again.

PR #438, opened 2017-07-17
Add basic SPI and GoPiGo3 support

iplay88keys:
This builds on the work done by JohannWeging for https://github.com/hybridgroup/gobot/pull/338. I have been able to implement the led and motor controls and have a framework set up for implementing additional controls.

No tests have really been written, but this provides a base for others to build on.

PR #448, opened 2017-08-25
First cut of the GoPiGo3 driver, lacking grove peripherals

ulisesflynn:
This builds on the collaboration with @brimstone and @iplay88keys from the GopherCon 2017 hack day. It makes use of the golang.org/x/exp/io/spi library for SPI support, and is based from GoPiGo3 python library. 

It's fully functioning, minus the grove peripherals, but be sure to run the GoPiGo3 shell script: https://www.dexterindustries.com/GoPiGo/get-started-with-the-gopigo3-raspberry-pi-robot/3-program-your-raspberry-pi-robot/python-programming-language/ before running this on the raspberry pi.


deadprogram:
Hi @ulisesflynn this looks really awesome! Could you please change it to be against the `dev` branch instead of `master`? Thank you!

deadprogram:
Thanks for changing the branch for merge, @ulisesflynn 

Last request I think that the file `drivers/spi/gopigo3_driver.go` should go into `platforms/gopigo/gopigo3_driver.go` instead since it is specific to the GoPiGo "platform", similar to the PR https://github.com/hybridgroup/gobot/pull/438 that @iplay88keys submitted. 

When we have an ssd1306 driver that supports the SPI interface, that would be an example of something that would go into `drivers\spi\ssd1306_driver.go`.

This PR has general purpose SPI communication on the raspi, as well as the GoPiGo implementation, so I think it is a great push in that direction. I'd love to have the platform move I mention above, what  does everyone think?

ulisesflynn:
Thanks Ron for the feedback. I did see putting the gopigo3 into a platform from the PR from @iplay88keys, but I was not sure at the which way to go.  However it does make sense to move it to a platform, and if everyone is good with it, I'll make the change.

examples/gopigo3.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	"fmt"
+	"time"
+
+	"gobot.io/x/gobot"
+	g "gobot.io/x/gobot/platforms/gopigo3"
+)
+
+func main() {
+	gopigo3Adaptor := g.NewAdaptor("/dev/spidev0.1", 0, 500000)
```
deadprogram: This example will need to be updated, right @ulisesflynn ?

platforms/gopigo3/README.md
```diff
@@ -0,0 +1,52 @@
...
+import (
+	"fmt"
+	"time"
+
+	"gobot.io/x/gobot"
+	g "gobot.io/x/gobot/platforms/gopigo3"
+)
+
+func main() {
+	gopigo3Adaptor := g.NewAdaptor("/dev/spidev0.1", 0, 500000)
```
deadprogram: Docs updated needed here, I think.

deadprogram:
I'm getting very excited about this!

deadprogram:
@ulisesflynn looks like this is ready! Anything else you want to add, or can I merge into `dev` now?

ulisesflynn:
Thanks @deadprogram I think this is good for now, and I've tested this with the gopigo3 hardware.

deadprogram:
I am now merging into `dev` branch. 

Thank you very much @ulisesflynn @brimstone and @iplay88keys for all of the code and efforts! This work both adds the GoPiGo3 platform, and also moves forward the overall goal of adding SPI interface to Gobot.

:1st_place_medal: :1st_place_medal: :1st_place_medal:

PR #452, opened 2017-09-26
Added grove support, and more gopigo3 examples

ulisesflynn:


deadprogram:
This looks great @ulisesflynn thank you! 

I noticed that the build failed, due to the test coverage dropping. Any chance of adding a few tests? :sun_with_face:

ulisesflynn:
Thanks @deadprogram! I've finally had time to add tests to driver. What do I need to add to pass the codecov/patch threshold?

deadprogram:
Thank you for adding some test coverage, @ulisesflynn 

The codecov "patch" test checks the % of test coverage of the new code. According to the stats, this PR has 66.52% of diff hit (target 84.87%), which I think means that 66.52% of the new code has code coverage, and codecov wants to see at least 84.87%. It is very strict.

Since the overall code coverage for the entire project is going up with this PR by 1.3% I'm not really concerned about those few lines. Additional test coverage could be added in a subsequent PR.

In other words, time to finally merge this into `dev` branch!

ulisesflynn:
Awesome, thank you for all your help with this PR! I'll try to add more testing to get better coverage in the next PR.

PR #455, opened 2017-10-07
Add dep, change how tests run in CI

### LINKED ISSUE #454: Use dep for dependency management
## What
I'm proposing to use [dep](https://github.com/golang/dep) as a tool to ensure a canonical dependency installation process for development.

## Why
It will make development far smoother and make reproducible builds much easier.

## Caveat
The `dep` tool requires Go 1.8 or higher. I believe that this is acceptable given that the Go team has committed to making this **the** tool for managing dependencies in Go software projects, but I admit I'm not fully aware of what concerns drive Gobot using 1.7 as the minimal supported version.

trevrosen:
Fixes #454

## What
This PR moves dependency management out of a manually maintained list in the `Makefile` and into dep artifacts (see issue #454). It also moves CI checks from in-lined bash in `.travis.yml` into `Makefile` targets and ensures that tests will exclude code in the `vendor/` directory. It also implements a variety of other build fixes and opinions as detailed below.

## Why
This will make development easier, less error-prone, and enable some dependency re-factoring that I've wanted to do for awhile. More broadly, the changes in this PR ensures that tests are run **the same way** in both development and CI.

## What Else
A number of improvements / alterations to the build process:

* `.travis.yml` now delegates everything to tasks defined in `Makefile`
* `Makefile`'s targets delegate to shell scripts in `ci/` dir where necessary
* Formatting check will fail the build if `gofmt -l` returns any files and print the problem paths
* All CGO-specific setup is now located in `test.sh`


trevrosen:
@deadprogram this is now (at long last!) ready for review. Once you bless, I'll squash merge so that my out-loud thinking isn't in our Git history... 😺

deadprogram:
Hi @trevrosen this really looks great, thank you so very much for working on it! Now merging.

PR #465, opened 2017-11-05
Change init payload sequence within jhd1313m1 driver Start() func.

sapiens-sapide:
+ code transposed from github.com/Seeed-Studio/Grove_LCD_RGB_Backlight/blob/master/rgb_lcd.cpp.

+ this fix ensures that the 2 lcd lines are addressable.


deadprogram:
Thank you for the corrections, @sapiens-sapide now merging.

PR #470, opened 2017-12-02
Adding stepper motor module

coding-yogi:
This module will allow an user to control a 4-wired stepper motor with a Single, Dual and Half stepping modes

### DIFF
--- a/drivers/gpio/stepper_driver.go
+++ b/drivers/gpio/stepper_driver.go
@@ -1,0 +1,230 @@
+package gpio
+
+import (
+	"errors"
+	"math"
+	"strings"
+	"sync"
+	"time"
+
+	"gobot.io/x/gobot"
+)
+
+type phase [][4]byte
+
+// StepperModes to decide on Phase and Stepping
+var StepperModes = struct {
+	SinglePhaseStepping [][4]byte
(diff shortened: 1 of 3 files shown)


deadprogram:
This is great @coding-yogi thank you.

Any chance you could also add a very basic example of it in use to `examples`? This also helps when we're preparing the docs.

Thanks!

coding-yogi:
Sure, let me add one

coding-yogi:
I am able to drive motor with raspi adaptor but not with gpg3, any pointers?
I see this error : **open /dev/spidev0.1: no such file or directory**.
 Here is my [code](https://gist.github.com/coding-yogi/41988b6928c5a32ac66a5c50af7d1035)

deadprogram:
The pin mapping/naming on the GoPiGo can be found here:

https://github.com/hybridgroup/gobot/blob/master/platforms/dexter/gopigo3/driver.go#L619

Seems like you would be better off with just a pure Raspi example. Even when using with the GoPiGo3, if you connected the pins to the stepper controller, you would just connect the Raspi pins, and use the GoPiGo3 grove ports for other devices.

deadprogram:
Thanks again for contributing @coding-yogi now merging.

Issue #471, opened 2017-12-02
Question: motor driver for GoPiGo platform


kpeu3i:
Does Gobot support motor driver for GoPiGo platform?
I can't find any examples how to control motors. 

Thanks in advance!

deadprogram:
Hi @kpeu3i there is support for the GoPiGo3 motors:

https://godoc.org/gobot.io/x/gobot/platforms/dexter/gopigo3#Driver.SetMotorDps
https://godoc.org/gobot.io/x/gobot/platforms/dexter/gopigo3#Driver.SetMotorLimits
https://godoc.org/gobot.io/x/gobot/platforms/dexter/gopigo3#Driver.SetMotorPosition
https://godoc.org/gobot.io/x/gobot/platforms/dexter/gopigo3#Driver.SetMotorPower

There does not appear to be a specific example using those API, but we do have a few other examples to use as reference for the GoPiGo3 in general:

https://github.com/hybridgroup/gobot/blob/master/examples/gopigo3_servo.go

Hope that helps!

kpeu3i:
Hi, @deadprogram! You've helped me a lot! it works :)

Here is a simple working example (motors will rotate forth):

```
work := func() {
    gopigoDriver.SetMotorPower(gopigo3.MOTOR_LEFT+gopigo3.MOTOR_RIGHT, 100)
}

robot := gobot.NewRobot("gopigo",
    []gobot.Connection{raspiAdaptor, dualshockAdaptor},
    []gobot.Device{gopigoDriver, servoDriver, motorDriver},
    work,
)

robot.Start()
```

kpeu3i:
It looks like method `SetMotorDps` doesn't work:

```
work := func() {
    err := gopigoDriver.SetMotorDps(gopigo3.MOTOR_LEFT, 180)
    if err != nil {
        app.workError <- err
    }
}
```

Nothing happens when I run it on my robot. Did I miss something?

deadprogram:
I'm not completely sure what that function is supposed to do. Paging Dr. @ulisesflynn for assistance.

kpeu3i:
I think this function is a port of [GoPiGo Python Driver](https://github.com/DexterInd/GoPiGo3/blob/master/Software/Python/gopigo3.py#L480).

```
Set the motor target speed in degrees per second
    Keyword arguments:
        port -- The motor port(s). MOTOR_LEFT and/or MOTOR_RIGHT.
        dps -- The target speed in degrees per second
```

[...]

PR #473, opened 2017-12-12
fixed set/get bug with motor dps

ulisesflynn:
Fixed bug found with #471 as well as for reading the motor status.


deadprogram:
Thanks @ulisesflynn now merging.

PR #478, opened 2018-01-11
Added basic driver for BH1750 (light sensor), board GY-302

conejoninja:
Driver for supporting the i²c BH1750 light sensor, usually a blue board with the text BH1750 on one side and GY-302 on the other side

### DIFF
--- a/drivers/i2c/bh1750_driver.go
+++ b/drivers/i2c/bh1750_driver.go
@@ -21,6 +21,8 @@ const (
 	BH1750_ONE_TIME_LOW_RES_MODE      = 0x23
 )
 
+// BH1750Driver is a driver for the BH1750 digital Ambient Light Sensor IC for I²C bus interface.
+//
 type BH1750Driver struct {
 	name       string
 	connector  Connector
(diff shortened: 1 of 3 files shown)


drivers/i2c/bh1750_driver.go
```diff
@@ -0,0 +1,109 @@
...
+		return err
+	}
+
+	return
+}
+
+// Halt returns true if devices is halted successfully
+func (h *BH1750Driver) Halt() (err error) { return }
+
+func (h *BH1750Driver) RawSensorData() (level int, err error) {
```
trevrosen: Comment string?

drivers/i2c/bh1750_driver.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+	if err != nil {
+		return
+	}
+	level = int(buf[0])<<8 | int(buf[1])
+
+	return
+}
+
+func (h *BH1750Driver) Lux() (lux int, err error) {
```
trevrosen: Comment string?

drivers/i2c/bh1750_driver.go
deadprogram: Agreed, we need for Godocs generation.

drivers/i2c/bh1750_driver.go
deadprogram: Same as other comment, needed for godocs.

conejoninja:
Thanks for the feedback, added the comments and the sensor to the readme.md list.
It should be ok now.

trevrosen:
Merging now that @deadprogram's feedback has been addressed.

PR #490, opened 2018-02-11
Initial support for TM1638 modules

conejoninja:
Initial support for TM1638 modules, sometimes marked as JY-LKM1638 or LED&KEY

It uses a SPI-like communication protocol on regular pins. You could chain up to 6 modules (to share clock and data signals), but they are treated as independent modules.

Each one of them has 8 7-segmet displays (with a dot), 8 leds and 8 buttons, unfortunately support for buttons is not available since it needs "input pullup" mode, currently not supported by gobot.

### DIFF
--- a/drivers/gpio/tm1638_driver.go
+++ b/drivers/gpio/tm1638_driver.go
@@ -1,0 +1,288 @@
+package gpio
+
+import (
+	"math"
+
+	"strings"
+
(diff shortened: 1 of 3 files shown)


deadprogram:
Very cool @conejoninja thank you very much for the contribution. Now merging.

PR #492, opened 2018-02-17
Added initial support for AIP1640 led driver

conejoninja:
It's used in Wemos D1 mini's matrix LED shield.


conejoninja:
Thanks for merging, that was fast :)

I'm not sure if I have any sensor/device left at home, maybe I should continue with the website (new sensors are in the example section, but not in the drivers menu). Is there any guideline to follow or something I should know?

PR #496, opened 2018-02-24
Initial support for MAX7219 & MAX7221 led drivers

conejoninja:
** Read before merging **
This is the driver for MAX 7219 / 7221 LED drivers, but I was only able to test it with MAX 7219 (don't have any 7221), but it's supposed to work with it too.

The official datasheets mentions that only the MAX7221 supports SPI (not the 7219), but most of the libraries available for Arduino use a mix of SPI and direct control of the SS/CS pin. I opted for direct control of pins instead of SPI because it's more versatile (works on TCPFirmata for example), but the speed might be slower than using SPI.

At this point, I'm not sure if I should rename the driver to max7219 only and make a SPI one for max7221, make a spi for both of them in replacement of this or make a spi driver in addition to this, so you will be able to use whatever suits you better (gpio or spi). I think this is fine for most cases.

Additionally, the driver is pretty simple as it's usually sold separately and the led arrangement (if in a matrix) is not always the same (it doesn't even has to be a matrix, it could be a 7-segment display).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,8 @@ sudo: required
 dist: trusty
 go_import_path: gobot.io/x/gobot
 go:
-   - 1.8.7
    - 1.9.4
+   - "1.10"
    - tip
 matrix:
    allow_failures:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,30 @@
+1.9.0
+---
+* **beaglebone**
+  * update pin naming, docs, and examples for the latest Debian OS releases
+* **opencv**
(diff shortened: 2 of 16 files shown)


trevrosen:
@deadprogram you have a ruling here on whether we tighten the name up before merge?

deadprogram:
@conejoninja @trevrosen I suggest this should indeed be 2 drivers:

- a gpio driver for `max7219`
- a spi driver for `max72xx`

Seems like this allows the most flexibility based on the desired hardware interface.

conejoninja:
great! will make the changes accordingly

conejoninja:
Continued this on #520 and made it clearer with just one commit

PR #506, opened 2018-04-01
Slightly improve the contributor instructions.

eric-s-raymond:
Make the step of checking out the dev branch explicit.


deadprogram:
Hi @eric-s-raymond all PRs need to be made against the `dev` branch, as mentioned in this doc here: https://github.com/hybridgroup/gobot/blob/master/CONTRIBUTING.md#general-guidelines

Also, please make all PR's atomic in that they only do a single thing. Both of your outstanding PRs have the same commits in them, which makes them very hard to review.

I think the simplest resolution is to close these PR's and then you can resubmit if you so choose.

Thank you!

PR #507, opened 2018-04-01
 Make the BB blink demo actually useful without special equipment.

eric-s-raymond:
Also, be explicit about its assumptions in the header comment and provide a textual indication
of when the LED should be toggling.

This should also replace the separate beaglebone_blink_usr_led.go


deadprogram:
Hi @eric-s-raymond as mentioned in my other comment, this outstanding PR has 2 commits in it which do 2 different things, which makes it very hard to review.

I will close this PR and then you can resubmit if you so choose.

Thank you!

PR #511, opened 2018-04-14
joystick: removed double release event

jig:
The `_release` event for joysticks is issued twice.

The `_press` event is immediately followed by a r`_release` event always. When the button is afterwards released the event is (properly now) re-issued.

This pull request removes the first `_release` event.


deadprogram:
Thanks @jig for the PR. Now merging.

PR #525, opened 2018-05-08
Add tflight hotas x controller

SMerrony:
Add the Thrustmaster T-Flight HOTAS X flight controller

(My first PR - hope I got things right...)

### DIFF
--- a/platforms/joystick/bin/scanner.go
+++ b/platforms/joystick/bin/scanner.go
@@ -42,7 +42,7 @@ func run() int {
 					t.Timestamp, t.Hat, t.Value)
 			case *sdl.JoyDeviceEvent:
 				if t.Type == sdl.JOYDEVICEADDED {
-					joysticks[int(t.Which)] = sdl.JoystickOpen(t.Which)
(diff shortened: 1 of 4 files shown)


deadprogram:
Hi @SMerrony thank you for this PR. 

The chance you made to `scanner.go` is incompatible with the latest SDL. This is the reason that the tests are failing on Travis. The reason they pass on AppVeyor (Windows) is that AppVeyor uses an SDL build.

I am going to make a change to exclude the scanner.go from the test run anyhow, since it is not intentional that it run as part of the tests.

Now merging, thanks again for the PR.

jadr2ddude:
@deadprogram just for reference, what version of SDL are you using that does not pass? I had run the build with version 2.0.8 on my local machine.

deadprogram:
We should probably install that version ourselves on Travis ala https://stackoverflow.com/questions/48733550/how-to-add-a-recent-version-of-libsdl2-to-my-travis-ci-build

deadprogram:
Also, we need to update to https://github.com/veandco/go-sdl2/releases/tag/v0.3

PR #528, opened 2018-05-12
Log with Logrus; allow for logger injection

### LINKED ISSUE #508: Override the default log with logrus?
Is there a method to override the default `import "log"`? It would be really handy to ship framework log messages off to an IoT agitator along with my app logging, rather than have them inaccessible out on the device. Centralized logging may be out of scope, but I thought I'd ask.

trevrosen:
Fixes #508

## What
* Adds Logrus to the build
* Replaces usage of `log` package from std lib with Logrus
* Adds `RegisterLogger` functions to `api` and `gobot` packages

## Why
* Requested in #508 
* Logrus has more features than std lib
* Using a `RegisterLogger` function allows for injecting a common logger

## Questions
* Does this solve the request in #508? (cc @FerventGeek)
* Should we leave the `examples` directory alone on the theory that it's best to show only Gobot packages and stdlib there? I opted to change b/c of the injection behavior's availability, but could see an argument for keeping it pure stdlib.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -37,18 +37,7 @@
 [[projects]]
   branch = "master"
   name = "github.com/go-ble/ble"
-  packages = [
-    ".",
-    "darwin",
-    "linux",
-    "linux/adv",
-    "linux/att",
-    "linux/gatt",
-    "linux/hci",
-    "linux/hci/cmd",
-    "linux/hci/evt",
-    "linux/hci/socket"
-  ]
+  packages = [".","darwin","linux","linux/adv","linux/att","linux/gatt","linux/hci","linux/hci/cmd","linux/hci/evt","linux/hci/socket"]
   revision = "788214691384e85e345bff9fd5eeb046f5983594"
 
 [[projects]]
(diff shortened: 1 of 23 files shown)


trevrosen:
@FerventGeek please have a look here when you get a chance and let me know if this is what you have  in mind or if you'd like to see further extension. Also: as we appear to have the same employer, feel free to get in touch IRL any time re Gobot. 😺

trevrosen:
@deadprogram the thought occurs that we could go a step further and formalize a method of log support at either the driver or platform level. I've got a couple ideas around what that could look like.

trevrosen:
@deadprogram any thoughts on whether we should continue w/ merging here? I'll update to fix the deps conflict if so. I'd also like to **squash merge** this one if possible due to my own vain desire to keep some of my derping around from the log and reduce noise :sim

trevrosen:
Closing. May re-open later.

ferventgeek:
This is awesome, thanks for the update. Logging refactoring is always tricky- this will really help with IoT projects and random robot images with no serial connector. I'll have time to test in early January.

trevrosen:
The PR is closed now after being open for comment for 7 months. I’m not satisfied with the change as it stood anyway. May revisit at some point in the future.

ferventgeek:
Ahh, well it happens. Would still like integrated exportable logs, but then again who doesn't. ;-) I'm all over the Arduino/ESP32 gizmo over the holiday anyway. ;-)

Issue #531, opened 2018-05-15
Tello Driver: Why are the FlightData fields not exported?


SMerrony:
Am I missing something, or is there a reason the fields of the FlightData struct are not exported?

deadprogram:
Hi @SMerrony that was strictly an oversight. We'd be grateful for a PR that exported all of them!

SMerrony:
PR #532 submitted :-)

deadprogram:
Thanks!

deadprogram:
Version 1.11.0 has just been released that includes this. Now closing, thank you @SMerrony

PR #532, opened 2018-05-15
Export the FlightData fields (see Issue #531)

SMerrony:
Make the individual FlightData fields visible to the program using gobot. 
Fixes Issue #531


deadprogram:
Wow, thank was quick @SMerrony thanks! Now merging...

Issue #533, opened 2018-05-15
Tello Driver: Change some FlightData fields to bool type


SMerrony:
It seems to me that many of the fields in the FlightData struct can contain int16 values of either 0 or 1.  Eg. all the fields extracted in ParseFlightData with an ``& 0x1``

Would it not be much more idiomatic to represent these as bool values in Go?

Happy to make this change if you think it worthwhile.

deadprogram:
Seems like a good idea to me!

deadprogram:
Version 1.11.0 has just been released that includes this. Now closing, thanks @SMerrony !

PR #534, opened 2018-05-15
Change several fields in FlightData struct from int16 to bool

SMerrony:
Change declaration of yes/no type fields in FlightData from int16 to bool and change corresponding parsing in ParseFlightData func.

Addresses Issue #533


deadprogram:
Thanks @SMerrony :helicopter: now merging.

Issue #536, opened 2018-05-17
Tello Driver: Support for more modes.


cbwang2016:
Such as Bounce Mode, Throw & Go

cbwang2016:
@deadprogram Do you have the protocol information for these modes?

deadprogram:
@cbwang2016 you should try to combine efforts with @SMerrony who also is working on this.

cbwang2016:
@deadprogram Yes. And personally I think you should release a new version now for all those minor bug fixes for tello.

PR #537, opened 2018-05-17
Tello: Driver - Add Bounce() and PalmLand() funcs and associated events

SMerrony:
Add two new dji/tello flight commands and their associated events.
Changed message and command definitions to better reflect their 16-bit nature and put all in hex with decimal values as comments.
Treating all message types as 16-bit allows a slight simplification to handleResponse().

Partly addresses issue #536

### DIFF
--- a/platforms/dji/tello/driver.go
+++ b/platforms/dji/tello/driver.go
@@ -13,6 +13,9 @@ import (
 )
 
 const (
+	// BounceEvent event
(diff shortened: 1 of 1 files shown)


deadprogram:
This is great, @SMerrony any chance of also adding an example that uses them?

deadprogram:
This is so very cool, I'm merging right now.

PR #538, opened 2018-05-18
Update driver.go

cbwang2016:
According to my reverse-engineering, it should be like this. You should double check this.


deadprogram:
Hi @cbwang2016 I cannot find where I originally obtained those values from, and I suspect your efforts are probably more correct than whatever mine were. Hence, merging. Thank you for the PR!

PR #539, opened 2018-05-18
Add "Throw & Go" support to Tello

cbwang2016:
Add Throw & Go support(Usage same as TakeOff() function)

### DIFF
--- a/platforms/dji/tello/driver.go
+++ b/platforms/dji/tello/driver.go
@@ -67,6 +67,7 @@ const (
 	takeoffCommand          = 0x54
 	landCommand             = 0x55
 	flipCommand             = 0x5c
+	throwtakeoffCommand     = 0x5d
 )
 
 // FlipType is used for the various flips supported by the Tello.
@@ -281,7 +282,7 @@ func (d *Driver) TakeOff() (err error) {
 
 // Throw & Go support 
 func (d *Driver) ThrowTakeOff() (err error) {
-	buf, _ := d.createPacket(0x5d, 0x48, 0)
+	buf, _ := d.createPacket(throwtakeoffCommand, 0x48, 0)
 	d.seq++
 	binary.Write(buf, binary.LittleEndian, d.seq)
 	binary.Write(buf, binary.LittleEndian, CalculateCRC16(buf.Bytes()))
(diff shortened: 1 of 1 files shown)


platforms/dji/tello/driver.go
```diff
@@ -279,6 +279,17 @@ func (d *Driver) TakeOff() (err error) {
 	return
 }
 
+// Throw & Go support 
+func (d *Driver) ThrowTakeOff() (err error) {
+	buf, _ := d.createPacket(0x5d, 0x48, 0)
```
deadprogram: You should probably create a constant for the command byte here, to match the other commands.

cbwang2016: Oh, sorry, so stupid of me. Can you help me to edit my code?

deadprogram: If you make a change to your branch, commit it, and then push to Github, it should update this PR.

SMerrony: Also, add the event and handle it...

deadprogram: Good point @SMerrony yes, that is also needed.

cbwang2016: @SMerrony There's no such event, I think. It's just FlightData.ThrowFlyTimer and receives with FlightDataEvent

deadprogram:
Just out of curiously, how do you use this? You call it then throw the drone up in the air? :helicopter: 

I just want to make sure that is correct before I try it... perhaps a note in the GoDoc comment for the method would be good.

cbwang2016:
Yep. 
1. Put your drone on your hands
2. Execute the ThrowTakeOff function, the propellers would rotates slowly
3. Throw your drone up in the air
You can try it on Windows using my app: https://github.com/cbwang2016/tello-control

deadprogram:
Oops, when I merged the PR from @SMerrony it added a conflict with this one. @cbwang2016 can you resolve, or do you need assist?

deadprogram:
Thank you. Now merging.

cbwang2016:
@deadprogram When will you release a new version of Gobot?

deadprogram:
@cbwang2016 we will likely have a Gobot release next week, to correspond with a parallel release of GoCV.

Issue #516, opened 2018-05-01
Test failures on 32 bits platform


maruel:
On an ODROID-C1+ running Ubuntu 16.04, I'm now able to run almost all tests;

First with some prerequisites:
```
sudo apt install libusb-dev libsdl2{,-image,-mixer,-ttf,-gfx}-dev
```

Then running the tests:
```
 ~/go/src/gobot.io/x/gobot$ ./xvfb.sh go test ./...
(successes elided)
# gobot.io/x/gobot/platforms/joystick/bin
platforms/joystick/bin/scanner.go:45:50: cannot use t.Which (type sdl.JoystickID) as type int in argument to sdl.JoystickOpen
# gobot.io/x/gobot/platforms/leap
platforms/leap/leap_motion_driver_test.go:92:45: constant 4729292670 overflows int
FAIL    gobot.io/x/gobot/platforms/leap [build failed]
--- FAIL: TestMqttAdaptorConnectWithAuthError (0.00s)
        gobottest.go:14: mqtt_adaptor_test.go:93: 1 error occurred:

                * Network Error : dial tcp [::1]:1883: connect: connection refused - "*multierror.Error", should equal,  1 error occurred:

                * Network Error : Unknown protocol - "*multierror.Error"
FAIL
FAIL    gobot.io/x/gobot/platforms/mqtt 0.038s
```

- platforms/leap/leap_motion_driver_test.go seems easy to fix, the test just fails on 32 bits platform.
- platforms/joystick/bin/scanner.go is probably related to the fact I installed an old sdl2 version (2.0.4), which prints a ton of warnings. Haven't investigated yet.
- platforms/mqtt/mqtt_adaptor_test.go is kinda weird, not sure why this test exists.

Where xvfb.sh is:
```
#/bin/sh
# Run under xvfb

set -eu

if (which xvfb-run>/dev/null); then
  xvfb-run \
    -a "--server-args=-screen 0 1280x800x24 -ac -nolisten tcp -dpi 98 +extension RANDR" \
    "$@"
else
  echo "Please 'sudo apt-get install xvfb' first"
  exit 1
fi
```

jadr2ddude:
For the platforms/leap/leap_motion_driver_test.go fail:

[...]

PR #541, opened 2018-05-19
add temporary fix to "TestMqttAdaptorConnectWithAuthError"

jadr2ddude:
this fixes the build failure which occurs on many systems (from issue #516)

### DIFF
--- a/platforms/mqtt/mqtt_adaptor_test.go
+++ b/platforms/mqtt/mqtt_adaptor_test.go
@@ -1,12 +1,10 @@
 package mqtt
 
 import (
-	"errors"
 	"fmt"
 	"strings"
 	"testing"
 
-	multierror "github.com/hashicorp/go-multierror"
 	"gobot.io/x/gobot"
 	"gobot.io/x/gobot/gobottest"
 )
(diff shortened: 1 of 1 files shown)


jadr2ddude:
I cannot see any particularly good solution. Paho appears to use a public testing server [see this](https://iot.eclipse.org/getting-started#sandboxes). I am not sure how using the testing server would or would not work.

jadr2ddude:
@deadprogram we could potentially reimplement the Client interface in the mqtt package, although that wouldnt help test the connect code. What do you think?

deadprogram:
Basically, that seems like the most pragmatic approach at the moment.

trevrosen:
Hi @jadr2ddude -- thanks for your interest in Gobot! This is the oldest PR currently open, so you've now become a part of my quest to get to PR Inbox Zero 😄 

Do we want to:

1. Declare an interface for the MTT driver in this PR? 
2. Merge as-is? 
3. Close?

 @deadprogram lemme know your preference here when you get a chance. Not hard to do the interface route, which is good to have anyway.

jadr2ddude:
If you want me to modify this, I might have a bit of time in the next few days

trevrosen:
@jadr2ddude any update here?

jadr2ddude:
No. I will be able to set up an MQTT interface this week.

trevrosen:
We can re-visit this with a patch for dependency injection, but I'm going to close this particular patch.

Thanks for the effort here, @jadr2ddude  -- happy to collaborate w/ you if you want to take on the dep injection bit here to make this test a little less janky.

PR #543, opened 2018-05-19
fix joystick code by casting to an int

jadr2ddude:
NOTE: Untested (I dont have an actual joystick). For issue #516.


deadprogram:
Hi @jadr2ddude we probably do not want this code here, as it then fails with the latest update to SDL. See https://github.com/hybridgroup/gobot/pull/525#issuecomment-387696758

Issue #549, opened 2018-05-20
throttle unused in Tello driver?


cbwang2016:
I searched the word "throttle" in file https://github.com/hybridgroup/gobot/blob/dev/platforms/dji/tello/driver.go
The word appears only twice:
`rx, ry, lx, ly, throttle float32`
and
`axis5 := int16(660.0*d.throttle + 1024.0)`
There's not even a function to set this "throttle".
Is this a bug?

deadprogram:
Hi @cbwang2016 it is really just an overlooked function. A PR that implemented would be welcomed.

PR #550, opened 2018-05-21
slow/fast mode switch function for DJI Tello

cbwang2016:
#549


deadprogram:
Thanks @cbwang2016 now merging.

PR #554, opened 2018-05-24
Fix Joystick Scanner

imathno:


deadprogram:
Please see https://github.com/hybridgroup/gobot/pull/543

PR #561, opened 2018-06-05
Added version constraint for periph

galenchurch:
This is a fix/workaround of failing build due to changes to the dependency "periph.io/x/periph".  It looks like this commit was responsible: https://github.com/google/periph/commit/9bf5daf1f1bc6edbcb1e8255cb524b2c400e500c

Replacing `int64` with the more "readable" type `physic.Frequency` has caused an error.  I'd imagine future releases will have to fix the type in the calls (or cast) but since that repo regularly has changes pushed to master I assumed the first step in a sustainable fix would be to add the version constraint to Gopkg.toml.

Signed-off-by: Galen Church <galenechurch@gmail.com>

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -27,28 +27,14 @@
 
 [[projects]]
   name = "github.com/eclipse/paho.mqtt.golang"
-  packages = [
-    ".",
-    "packets"
-  ]
+  packages = [".","packets"]
   revision = "36d01c2b4cbeb3d2a12063e4880ce30800af9560"
   version = "v1.1.1"
 
 [[projects]]
   branch = "master"
   name = "github.com/go-ble/ble"
-  packages = [
-    ".",
-    "darwin",
-    "linux",
-    "linux/adv",
-    "linux/att",
-    "linux/gatt",
-    "linux/hci",
-    "linux/hci/cmd",
-    "linux/hci/evt",
-    "linux/hci/socket"
-  ]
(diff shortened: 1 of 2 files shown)


deadprogram:
Hi @galenchurch thank you for this PR. Don't we also need to update the `Gopkg.lock` file as well?

galenchurch:
Its up to you if you want me to include it in the PR. `Gopkg.lock` file is auto-updated and overwritten when `dep ensure` is run.  When I ran it there were some formatting changes to it (it removed all the line breaks on the arrays) so I left it out of the commit.  I can re-submit the PR with my updated version included if you would like? Just let me know if style is important ;)

deadprogram:
Question for @trevrosen should the Gopkg.lock be in the repo, or just built from `dep ensure` as mentioned by @galenchurch

trevrosen:
Yes pls update. That behavior with line breaks is due to a recent change in how dep formats lock files and is fine to commit.

galenchurch:
Ok, thanks...should be all set.

trevrosen:
@deadprogram not understanding this codecov diff, but if we're fine ignoring that, I'm good to merge.

deadprogram:
Something went wrong there on codecov, not too sure. I sent a support request already. Seems good to me!

deadprogram:
Merging this PR since test coverage report is not correct.

Issue #568, opened 2018-06-19
MCP23017 driver and Raspberry Pi - Read Buffer not correct


barrowkwan:
I am trying to use MCP23017 with Raspberry PI to turn PortA pin 0-7 ON ( ie write "1" to port A , pin 0-7).  However,  the driver didn't keep already "ON" pin to stay "ON".  Basically it can only keep one and only one of the pin 0-7 "ON".  There is no way to turn more than one pin ON at the same time.

After some debugging, it turns out the "read" function in mcp23017_driver.go return the wrong buffer.  "IODIR" supposed to be in buf[20] but it was in buf[19].

I don't know if this is a problem specific to RPi.  I have tried both RPi2 and RPi3.   In order to fix this, I have to change "read" function to return buf[register-1] instead.

	if register != 0 {
		return buf[register-1], nil
	} else {
		return buf[register], nil
	}

deadprogram:
@ulisesflynn any comments on this?

ulisesflynn:
@barrowkwan I am looking into this now, I'll get back with you by tomorrow.

deadprogram:
Hi @barrowkwan there is a PR from @ulisesflynn that appears should fix things that is located here:
https://github.com/hybridgroup/gobot/pull/569 

Are you able to test it out and ensure it solves your specific issues?

barrowkwan:
Yes that fixed it.  thanks!

barrowkwan:
Hi
  I think this clean up has introduce another bug.  I used mcp23017 to drive a 16 relay board, before the fix, I can turn the relay on or off ( I have to use the PinMode method to set the pin input/output mode ).  Right now the PinMode function has gone and I can't change the PIN's mode. 

After looking at the code, I think the problem is with the WriteGPIO.  We should be able to set 1 or 0 to the pin ( ie on or off ).  However, if we try to send a value "1" ( ie write 1 to the output pin, it will change the pin to be input instead of output ).

barrowkwan:
The problem is with the WriteGPIO function.  It should always use clearBit and write to IODIR.  so instead of

```

ulisesflynn:
I'll look at this tomorrow and see what's going on, I apologize for my slow response

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

ulisesflynn:
@barrowkwan I've pushed up a PR for the writegpio/readgpio code, let me know if this fixes it.

ulisesflynn:
[...]

PR #569, opened 2018-06-21
fixed bug in read/write gpio code, and cleaned up some comments

ulisesflynn:
This fixes bugs found in issue: https://github.com/hybridgroup/gobot/issues/568 as well as some general cleanup of the library.

### DIFF
--- a/drivers/i2c/mcp23017_driver_test.go
+++ b/drivers/i2c/mcp23017_driver_test.go
@@ -11,9 +11,9 @@ import (
 	"gobot.io/x/gobot/gobottest"
 )
 
+var _ gobot.Driver = (*MCP23017Driver)(nil)
 var (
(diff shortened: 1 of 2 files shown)


drivers/i2c/mcp23017_driver_test.go
```diff
@@ -11,18 +11,18 @@ import (
 	"gobot.io/x/gobot/gobottest"
 )
 
-var _ gobot.Driver = (*MCP23017Driver)(nil)
```
deadprogram: I'd leave this line alone since it is actually a test that ensures that the driver supports the `gobot.Driver` interface, or else it will not compile.

ulisesflynn: Thank you for reviewing the changes, I've added that line back to the test file.

deadprogram:
Thanks @ulisesflynn now merging.

PR #578, opened 2018-07-03
Bump periph.io/x/periph to v3.0.0

maruel:
There was one breaking change in the spi.Port and spi.Conn interfaces.

As I ran "dep ensure" with dep v0.4.1, it decided to reformat the Gopkg.lock
file. :/


deadprogram:
Thanks @maruel I was just going to look into doing this! :racehorse: Now merging.

PR #582, opened 2018-07-09
Remove scaling of user input and add comments to SetPWMFreq function in PCA9685

ssnover95:
Hello,

I was working with this driver today and noticed that there was a random prescaler on my argument in the source. I checked this when I was finding my frequency was much lower than expected. I imagine it was accounting for some error in the oscillator since there was no comment explaining what it was, but I don't feel that that should be in a driver, it should be up to the user to account for that. Without the prescaling I had better accuracy to my frequency.

I also added some comments as I wanted to make sure I hadn't missed something about the factor of 0.9 so I went through the datasheet. These were just comments that I would've wanted while reading through, so I thought I would add them now.

Cheers!

### DIFF
--- a/drivers/i2c/pca9685_driver.go
+++ b/drivers/i2c/pca9685_driver.go
@@ -104,12 +104,14 @@ func (p *PCA9685Driver) SetPWM(channel int, on uint16, off uint16) (err error) {
 
 // SetPWMFreq sets the PWM frequency in Hz
 func (p *PCA9685Driver) SetPWMFreq(freq float32) error {
-	freq *= 0.9
-
+	// IC oscillator frequency is 25 MHz
 	var prescalevel float32 = 25000000
+	// Find frequency of PWM waveform
 	prescalevel /= 4096
+	// Ratio between desired frequency and maximum
 	prescalevel /= freq
 	prescalevel -= 1
+	// Round value to nearest whole
 	prescale := byte(prescalevel + 0.5)
 
 	if _, err := p.connection.Write([]byte{byte(PCA9685_MODE1)}); err != nil {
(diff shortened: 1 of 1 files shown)


deadprogram:
Hello @ssnover95 I think you've got the right idea here. The reason for my original implementation was this: https://github.com/adafruit/Adafruit-PWM-Servo-Driver-Library/issues/11

More recently the Adafruit libraryhas this issue https://github.com/adafruit/Adafruit-PWM-Servo-Driver-Library/issues/40 which seems much closer to your implementation.

What do you think?

ssnover95:
The discussion in issue 40 resolves around the rounding because they're using integer math instead of float math. This is currently accounted for in the driver here with the addition of 0.5 which has the effect of mathematically rounding when converted to an integer instead of the normal "floor" function when a float is converted to integer.

There's just going to be variation in the oscillator frequency due to the environment and manufacturing variation. One possible solution would be to let the user specify a multiplier optionally which would be used to scale the nominal frequency, but the user would still need to measure it first. For the time being, removing the 0.9 scaling factor will sufficiently accuracy of the final output frequency.

deadprogram:
Sounds good @ssnover95 thanks for the contribution. Now merging.

photodude:
@ssnover95 @deadprogram FYI the Adafruit-PWM-Servo-Driver-Library now includes a method for "setting" the internal oscillator frequency which greatly improves the ability to get the desired PWM frequency from these PCA9685 chips. See https://github.com/adafruit/Adafruit-PWM-Servo-Driver-Library/commit/d1f63cc9f463e69ebc38d3283d51f6f7798ab437

Technically this new function doesn't set the frequency of the chip, but sets the value of a variable the library uses to calculate other things like the PWM frequency. Got to love how inconsistent these PCA9685 chips are in delivering an expected PWM frequency (also got to love the utilitarian ability to deliver lots of PWM signals with these chips).

Hopefully, this information is helpful to you.

PR #583, opened 2018-07-13
[WIP] Tello driver refactor

oliverpool:
Hi,

gobot looks so cool, that I had to get a Tello drone to play with it :-)

I got started in no-time thanks to the documentation and example: kudos for that!

I've looked at the source, and I think that the code could be refactored to improve its readability and its testability (now that you figured out the hard part of the messages content and format).

This is the work that I've done this far (mainly `handleResponse` testing). Before going further, I would like to know if you think that it is a good idea! (if not, I would rather know it before I invest too much time in it ;-)

### DIFF
--- a/platforms/dji/tello/driver.go
+++ b/platforms/dji/tello/driver.go
@@ -183,7 +183,7 @@ type WifiData struct {
 type Driver struct {
 	name           string
 	reqAddr        string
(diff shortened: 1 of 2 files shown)


deadprogram:
Hi @oliverpool this looks really great, thank you for working on it. Also, thank you for starting to add test coverage, which is really needed. Please continue! Now merging this PR.

PR #592, opened 2018-08-12
tello: add example with keyboard

abtris:


trevrosen:
Thanks for the PR!

FYI this Travis fail looks to be an ephemeral build fail that needs addressing at the build level. Landing this.

PR #594, opened 2018-08-15
raspi: export PiBlasterPeriod in Adaptor

starryalley:
By setting PiBlasterPeriod to a raspi.Adaptor, the PWM pin will now reflect the correct duty cycle to /dev/pi-blaster. The default value is still 10000000(ns), which is the default one in pi-blaster too.

In addition to exporting PiBlasterPeriod in raspi.Adaptor, I also have to create a pointer in PWMPin structure to point to the Adaptor's PiBlasterPeriod, in order for Period() and SetDutyCycle() in all PWMPin to reference to the same value, which should be the same throughout runtime. This hence changes the NewPWMPin() interface.

Tests updated accordingly.

Still new to golang. Let me know if something is unexpected. Thanks.

### DIFF
--- a/platforms/raspi/pwm_pin.go
+++ b/platforms/raspi/pwm_pin.go
@@ -9,13 +9,12 @@ import (
 	"gobot.io/x/gobot/sysfs"
 )
 
-const piBlasterPeriod = 10000000
-
 // PWMPin is the Raspberry Pi implementation of the PWMPinner interface.
 // It uses Pi Blaster.
 type PWMPin struct {
-	pin string
-	dc  uint32
+	pin    string
+	dc     uint32
+	period uint32
 }
 
 // NewPwmPin returns a new PWMPin
@@ -51,11 +50,19 @@ func (p *PWMPin) InvertPolarity(invert bool) (err error) {
 
 // Period returns the current PWM period for pin
 func (p *PWMPin) Period() (period uint32, err error) {
-	return piBlasterPeriod, nil
+	if p.period == 0 {
+		return p.period, errors.New("Raspi PWM pin period not set")
+	}
+
+	return p.period, nil
 }
 
-// SetPeriod does not do anything when using PiBlaster
+// SetPeriod uses PiBlaster setting and cannot be changed once set
 func (p *PWMPin) SetPeriod(period uint32) (err error) {
+	if p.period != 0 {
+		return errors.New("Cannot set the period of individual PWM pins on Raspi")
+	}
+	p.period = period
 	return nil
 }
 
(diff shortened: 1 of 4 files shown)


platforms/raspi/pwm_pin.go
```diff
@@ -9,19 +9,19 @@ import (
...
-	pin string
-	dc  uint32
+	pin    string
+	dc     uint32
+	period *uint32
 }
 
 // NewPwmPin returns a new PWMPin
-func NewPWMPin(pin string) *PWMPin {
+func NewPWMPin(pin string, period *uint32) *PWMPin {
```
deadprogram: The `NewPWMPin()` function signature is the same across all the Adaptors that support it, so you probably should not add another param here.

starryalley: Thanks. I'll try other way around it.

starryalley:
Yes pi-blaster uses a fixed period and affects all pins on that board. I thought SetPeriod() should actually change the period so I didn't choose to use it. However since pi-blaster doesn't support this we can then use SetPeriod()? I'll update the PR.

deadprogram:
Here is what I would do:

First,add a new member `PiBlasterPeriod` to the Raspi's Adaptor struct who's default value is `10000000`: https://github.com/hybridgroup/gobot/blob/master/platforms/raspi/raspi_adaptor.go#L24

This allows you to create your adaptor like this:

```go
r := raspi.NewAdaptor()
r.PiBlasterPeriod = 20000000
```

Next, modify the Raspi's `PWMPin` struct to include a member `period uint32`.
https://github.com/hybridgroup/gobot/blob/master/platforms/raspi/pwm_pin.go#L16

Then modify the `SetPeriod()` method
https://github.com/hybridgroup/gobot/blob/master/platforms/raspi/pwm_pin.go#L58

With something like this:

```go
// SetPeriod uses PiBlaster setting, so it cannot be changed once set.
func (p *PWMPin) SetPeriod(period uint32) (err error) {
    if p.period != 0 {
        return errors.New("Cannot set the period of individual PWM pins on Raspi")
    }
    p.period = period
    return nil
}
```

Then, change the Raspi's Adaptor's `PWMPin()` method, so that right after it creates the `NewPWMPin()`, it calls

```go
r.pwmPins[i].SetPeriod(r.PiBlasterPeriod)
```

https://github.com/hybridgroup/gobot/blob/master/platforms/raspi/raspi_adaptor.go#L263

In theory, this should give you exactly what you are looking for.

starryalley:
@deadprogram The proposed way looks great! I'll give it a try and update. Thanks!

deadprogram:
Hi @starryalley thanks for the PR updates. I changed the branch from `master` to `dev` as mentioned here https://github.com/hybridgroup/gobot/blob/master/CONTRIBUTING.md#general-guidelines

Now merging, thanks again!

starryalley:
Didn't notice I should push to `dev`. Thanks!

PR #601, opened 2018-08-30
Automation model

trevrosen:
## What
An evolution in the way Gobot manages automation, this PR introduces the concept of `RobotWork`, managed in a `RobotWorkRegistry`. This construct exists on the `Robot` type and is accessed by new `Every` and `After` methods implemented directly on a `Robot`. 

In contrast to the existing `Every` and `After` functions, these `Robot` methods take a `context.Context` as well as a duration and a function. Calling `Robot.Every()` and `Robot.After()` will create new `RobotWork` instances, place them in a goroutine-safe registry, and set up a cancellable context.

## Why
This allows us to extend the semantics of automation in Gobot by placing the responsibility for managing discreet work unit on the `Robot` itself, along side attendant `Driver` and `Connection` types, making for easier extension of these related concerns later.
* Repeating or future-scheduled workloads can be cancelled externally and synchronized together
* `Robot.Every` tracks successful iterations of each work function with a monotonic counter
* Using `context.Context` allows for the usage of powerful patterns in orchestrating work units

## Caveats/Questions
* I *think* I've got the goroutine-safety stuff covered, but would like a bit more eyes on them
* Have I missed any tests?

### DIFF
--- a/robot_work.go
+++ b/robot_work.go
@@ -1,0 +1,170 @@
+// RobotWork and the RobotWork registry represent units of executing computation
+// managed at the Robot level. Unlike the utility functions gobot.After and gobot.Every,
+// RobotWork units require a context.Context, and can be cancelled externally by calling code.
+package gobot
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"sync"
+
+	"github.com/gobuffalo/uuid"
+)
+
+// RobotWorkRegistry contains all the work units registered on a Robot
+type RobotWorkRegistry struct {
+	sync.RWMutex
+
+	r map[string]*RobotWork
+}
+
+const (
+	EveryWorkKind = "every"
+	AfterWorkKind = "after"
+)
+
+// RobotWork represents a unit of work (in the form of an arbitrary Go function)
+// to be done once or on a recurring basis. It encapsulations notions of duration,
+// context, count of successful runs, etc.
+type RobotWork struct {
+	id         uuid.UUID
+	kind       string
+	tickCount  int
+	ctx        context.Context
+	cancelFunc context.CancelFunc
+	function   func()
+	ticker     *time.Ticker
+	duration   time.Duration
+}
+
+// ID returns the UUID of the RobotWork
+func (rw *RobotWork) ID() uuid.UUID {
+	return rw.id
+}
+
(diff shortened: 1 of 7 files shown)


trevrosen:
@deadprogram any objection to me just landing this? 😄

deadprogram:
I promise I will review by tomorrow, @trevrosen

robot_work.go
```diff
@@ -0,0 +1,200 @@
...
+	return fmt.Sprintf(format, rw.id, rw.kind, rw.tickCount)
+}
+
+// WorkRegistry returns the Robot's WorkRegistry
+func (r *Robot) WorkRegistry() *RobotWorkRegistry {
+	return r.workRegistry
+}
+
+// Every calls the given function for every tick of the provided duration.
+func (r *Robot) Every(ctx context.Context, d time.Duration, f func()) *RobotWork {
```
deadprogram: Just wondering if this method should be:

```
func (r *Robot) Every(d time.Duration, ctx context.Context, f func()) *RobotWork
```

The idea being that the signature would be a little closer to the `gobot.Every()` function.

robot_work.go
```diff
@@ -0,0 +1,200 @@
...
+				rw.tickCount++
+			}
+		}
+		r.WorkEveryWaitGroup.Done()
+	}()
+	return rw
+}
+
+// After calls the given function after the provided duration has elapsed
+func (r *Robot) After(ctx context.Context, d time.Duration, f func()) *RobotWork {
```
deadprogram: Same comment applies to `After()` as did to `Every()`

robot_work.go
trevrosen: I'm not dead-set against that, but in most Go code that takes a context, the context is usually passed as the first arg by convention. This still leaves the `time.Duration` and the `func()` next to each other -- potentially it's *more* confusing to have those two split by the new var in the signature, no?

deadprogram: Shows how much I know about context. OK, then, let's just carry on with it.

trevrosen: 😺 I tried to basically follow how it works in the stdlib. Here's an example: https://tip.golang.org/pkg/net/#Dialer.DialContext

deadprogram:
Good work on this @trevrosen now merging.

PR #615, opened 2018-09-13
NATS Updates

wallyqs:
Hi, just updating the examples in the README and also the import to the NATS client that has moved to github.com/nats-io/go-nats

Thanks!


wallyqs:
Thanks for the merge! I'll submit to dev branch next time, sorry I missed this from the contrib docs...

PR #616, opened 2018-09-14
Add SparkFun’s EasyDriver (and BigEasyDriver) Stepper Motor Driver

brendanstennett:
Signed-off-by: Brendan Stennett <brendan@thinkdataworks.com>

This comes complete with Run/Stop/Move/Step/Enable/Disable/Sleep/Wake

I have tested it with my EasyDriver and the one stepper motor I have.  Will test with more as I obtain them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,6 +231,7 @@ a shared set of drivers provided using the `gobot/drivers/gpio` package:
 	- Button
 	- Buzzer
 	- Direct Pin
+	- EasyDriver
 	- Grove Button
 	- Grove Buzzer
 	- Grove LED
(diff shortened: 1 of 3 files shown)


brendanstennett:
And let me know if you would prefer to name this off the chip (A3967) instead of this board? I figured the board because it does seem to be quite popular but the chip would be more general.

brendanstennett:
I added a couple extra tests to try and get coverage to a point that is passable.  They are pretty thin though so I'm not sure if you really want these types of tests included.  I have them as a separate commit for your review but I can rebase and squash when ready.

deadprogram:
Thannks @brendanstennett for the PR. Looks good, thanks for adding the test coverage. Now merging.

Issue #617, opened 2018-09-14
drivers/gpio: SA4000 warning in max7219_driver.go


quasilyte:
Both gocritic and megacheck (staticcheck) find these code lines suspicious:

https://github.com/hybridgroup/gobot/blob/d7818a756eb3c2b275ddf9138e0e529dfb87f5dd/drivers/gpio/max7219_driver.go#L89
https://github.com/hybridgroup/gobot/blob/d7818a756eb3c2b275ddf9138e0e529dfb87f5dd/drivers/gpio/max7219_driver.go#L65

`&` is used with same RHS and LHS operands.

Should probably be simplified as `x&x`=>`x`, but as it looks somewhat confusing, I've opened an issue instead of sending a PR (maybe there is some IoT-related domain convention so `x&x` makes sense here).

deadprogram:
There is now code in the `dev` branch thanks to @conejoninja so I will leave this open until the next release. Thanks everyone!

deadprogram:
This code was release in v1.13.0 thank you! Now closing.

PR #618, opened 2018-09-16
Simplify code as suggested in #617

conejoninja:
I can't remember why, but it could be simplified


deadprogram:
Thanks @conejoninja now merging.

PR #627, opened 2018-10-02
tello: add direct vector access

silkeh:
This PR adds direct access to the flight data vector, along with some syntaxis sugar.
The new functions perform a translation from a regular flight vector with three translation axes and one rotational axis (see `Vector()`) to the one required by the Tello, which is not very clear.

### DIFF
--- a/platforms/dji/tello/driver.go
+++ b/platforms/dji/tello/driver.go
@@ -420,6 +420,103 @@ func (d *Driver) Rate() (err error) {
 	return
 }
 
+// bound is a naive implementation that returns the smaller of x or y.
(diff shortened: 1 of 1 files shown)


trevrosen:
@silkeh I'm OK with merging this, but I'd love a bit more context about what's driving the change here and what this is enabling. That info could be usefully captured as documentation on the `Vector()` method. Apologies for getting to this so late, and thanks for your interest in Gobot!

trevrosen:
@silkeh thanks for the patch -- sorry the merge has taken a long time 😢 

@deadprogram This is purely additive and I see no reason not to merge it, so I'm going ahead.

PR #630, opened 2018-10-09
Add CCS811 I2C Driver

kml183:
# Overview
This code adds a new I2C driver for the CCS811 gas sensor. The documentation for the sensor can be found [here](https://cdn-learn.adafruit.com/assets/assets/000/044/636/original/CCS811_DS000459_2-00-1098798.pdf?1501602769). The chip can sense tvoc(Total Volatile Organic Compounds) and co2 concentrations. 

# Proof of Performance
The following image contains the output you would see from running the code in `examples/raspi_ccs811.go`. 
screenshot from 2018-10-08 23-18-32

# Testing
I was using this [breakout board](https://www.adafruit.com/product/3566). The documentation for the PCB can be found [here](https://learn.adafruit.com/adafruit-ccs811-air-quality-sensor/downloads). Note that the temperature calculation can only be done if an resistor is installed between pins 4 and 8. I tested the driver on the raspberry pi 3+ and the raspberry pi zero.

### DIFF
--- a/drivers/i2c/ccs811_driver.go
+++ b/drivers/i2c/ccs811_driver.go
@@ -1,0 +1,349 @@
+package i2c
+
+import (
+	"fmt"
+	"math"
+	"time"
+
+	"gobot.io/x/gobot"
+)
+
+// CCS811DriveMode type
+type CCS811DriveMode uint8
+
+// Operating modes which dictate how often measurements are being made. If 0x00 is used as an operating mode, measurements will be disabled
+const (
+	CCS811DriveModeIdle  CCS811DriveMode = 0x00
+	CCS811DriveMode1Sec                  = 0x01
+	CCS811DriveMode10Sec                 = 0x02
+	CCS811DriveMode60Sec                 = 0x03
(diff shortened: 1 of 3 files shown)


drivers/i2c/ccs811_driver.go
```diff
@@ -0,0 +1,349 @@
...
+//GetTemperature returns the device temperature in celcius.
+//If you do not have an NTC resistor installed, this function should not be called
+func (d *CCS811Driver) GetTemperature() (float32, error) {
+
+	buf, err := d.read(ccs811RegNtc, 4)
+	if err != nil {
+		return 0, err
+	}
+
+	vref := ((uint16(buf[0]) << 8) | uint16(buf[1]))
```
kml183: I grabbed the needed calculation from [here](https://github.com/adafruit/Adafruit_CircuitPython_CCS811/blob/master/adafruit_ccs811.py#L165)

deadprogram:
This is great @kml183 thank you very much for the contribution. Now merging...

PR #631, opened 2018-10-09
Fix slice bounds out of range panic when stopping Tello drone

eleniums:
Most of the time when stopping my DJI Tello drone, I was getting this panic:
```
panic: runtime error: slice bounds out of range

goroutine 50 [running]:
gobot.io/x/gobot/platforms/dji/tello.(*Driver).processVideo.func1(0xc4200d2000)
	/Users/snelson/go/src/gobot.io/x/gobot/platforms/dji/tello/driver.go:819 +0x24e
created by gobot.io/x/gobot/platforms/dji/tello.(*Driver).processVideo
	/Users/snelson/go/src/gobot.io/x/gobot/platforms/dji/tello/driver.go:815 +0xe3
exit status 2
```
I noticed the driver was trying to publish a VideoFrameEvent before checking for an error from the read UDP operation. This commit moves the error check to immediately after the read operation and continues if an error occurred. This avoids trying to slice an empty buffer and dodges a panic.

### DIFF
--- a/platforms/dji/tello/driver.go
+++ b/platforms/dji/tello/driver.go
@@ -816,11 +816,12 @@ func (d *Driver) processVideo() error {
 		for {
 			buf := make([]byte, 2048)
 			n, _, err := d.videoConn.ReadFromUDP(buf)
-			d.Publish(d.Event(VideoFrameEvent), buf[2:n])
-
 			if err != nil {
 				fmt.Println("Error: ", err)
+				continue
 			}
+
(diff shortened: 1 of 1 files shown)


eleniums:
Do I need to increase code coverage in order for this pull request to be addressed and accepted?

deadprogram:
Hi, @eleniums thank you very much for this PR. Test coverage would be nice, certainly, but the coverage in this platform is not very high to begin with. In any case, now merging, as I need this myself! Thanks again.

PR #633, opened 2018-10-12
Fix DJI Tello video example

eleniums:
This change allows mplayer to keep up with the video frame events. With the fps set to 25, the video frame events were stacking up and the video was lagging horribly behind.


trevrosen:
@eleniums Thanks for the contribution! Finally reviewing here in a minute... 😸

PR #645, opened 2018-12-17
Allow setting QoS on MTT adaptor

### LINKED ISSUE #644: Set QoS for MQTT Publish/Subscribe
It would be nice to be able to set the QoS when subscribing/publishing to events. The config stream in google cloud iot-core for instance recommends a QoS of 1.

trevrosen:
Fixes #644 

## What
Add an option for setting the QoS value sent with MTT messages in publish/subscribe events

## Why
Requested in #644 and also extends the value of the adaptor generally.


deadprogram:
Excellent change, thanks for doing it so quickly. Now merging.

PR #646, opened 2018-12-25
Update Leap Motion support to API v6

samkass:
The difference from v3 to v6 of the Leap Motion REST API wasn't that big of a change to the gobot Platform module, but makes a huge difference to the amount of information you can get from the Leap Motion. This is my first go code, but I tried to match style and idiosyncrasies. I tested it with my own Leap Motion on my Mac and the new information seems to be available.


samkass:
I realized there was an existing example of HandIDs, so I changed HandIds and PointableIds from Gesture to HandIDs and PointableIDs respectively. I also updated the unit test to match the new sample data.

samkass:
Ugh. I just realized I made this pull request against master. I will open a new pull request against dev. Sorry for all the noise.

PR #647, opened 2018-12-25
Update gobot leap platform to support Leap Motion API v6

samkass:
This pull request updates the leap platform module from Leap Motion REST API v3 to v6. The changes to the code are rather small, but the additional information from the Leap Motion with the new API is substantial. This is my first go code, but I tried to match idiosyncracies and style.

This change does include 3 breaking changes. In Gesture, Hands is now HandIDs, and Pointables is now PointableIDs. In InteractionBox, center is now an array of float64 (as per the API spec) instead of int. No breaking changes were made to Pointable or Hand (only additional fields). I suspect most users of the Leap Motion API were not using the fields in question, but if so they'll have to update their code.

New features include the ability to get the arm position vectors, the finger bone positions, the finger type (thumb, etc), the finger size, as well as additional details about gestures.


trevrosen:
@deadprogram I like this patch and am wondering how you want to handle version implications.

trevrosen:
@samkass thanks for this patch -- I'm going to merge this as-is and we'll figure out version implications at release time. Sorry for the delay in merging!

PR #650, opened 2019-01-03
Add some new MQTT adaptor functions with QOS

trevrosen:
#644 

## What
Implements some discussion on #644 around extending the methods on the adaptor

## Why
Allow calling code to set per-event publish and subscribe QOS values, use the `errors` package to return a richer error when the client isn't available.


deadprogram:
This looks great @trevrosen exactly what I was hoping for. Now merging. Thank you!

PR #652, opened 2019-01-20
minor updates to opencv README

NikolaeVarius:
`source $GOPATH/src/gocv.io/x/gocv/env.sh`

returns
```
This script is no longer necessary and has been deprecated.
See the Custom Environment section of the README if you need to customize your environment.
```
Also GoCV installed openCV4 which should be pointed out


trevrosen:
Thanks for the patch -- this will be helpful to folks in the future for sure.

PR #653, opened 2019-01-24
Added some new features to the sphero ollie, bb-8 and sprkplus

Rearth:
Signed-off-by: Rearth <davidus2010@gmail.com>

This adds missing functionality to the sphero ollie, bb-8 and sprkplus. Those features are already included in the "normal" sphero, but since the newer spheros use BLE, they haven't been able to use the locator and data-streaming of the sphero driver. 
This PR fixes that by adding the methods to set a datastreamingconfig and reading the locator data. Also, the "datastreaming" event has been added, to listen to the datastreaming events.
Unit tests for the new behaviors are included.

### DIFF
--- a/platforms/sphero/ollie/ollie_driver.go
+++ b/platforms/sphero/ollie/ollie_driver.go
@@ -19,7 +19,10 @@ type Driver struct {
 	seq               uint8
 	mtx               sync.Mutex
 	collisionResponse []uint8
-	packetChannel     chan *packet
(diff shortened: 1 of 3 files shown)


trevrosen:
@Rearth hey thanks for your patch -- I will not have time to review until later in the week.

@deadprogram have you been able to take a look?

trevrosen:
Thanks for the work here @Rearth -- glad to see this patch!

PR #660, opened 2019-03-17
Fix Gopkg.lock is out of sync error

dlisin:
When running `dep ensure` on local machine - following warning message is displayed:
```
Warning: the following project(s) have [[constraint]] stanzas in Gopkg.toml:

  ✗  github.com/nats-io/nats

However, these projects are not direct dependencies of the current project:
they are not imported in any .go files, nor are they in the 'required' list in
Gopkg.toml. Dep only applies [[constraint]] rules to direct dependencies, so
these rules will have no effect.

Either import/require packages from these projects so that they become direct
dependencies, or convert each [[constraint]] to an [[override]] to enforce rules
on these projects, if they happen to be transitive dependencies.

# Gopkg.lock is out of sync with Gopkg.toml and project imports:
github.com/nats-io/go-nats: imported or required, but missing from Gopkg.lock's input-imports
github.com/nats-io/nats: in Gopkg.lock's input-imports, but neither imported nor required
```

This PR fixes this

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -151,17 +151,10 @@
   digest = "1:f04a78a43f55f089c919beee8ec4a1495dee1bd271548da2cb44bf44699a6a61"
   name = "github.com/nats-io/go-nats"
   packages = [
+    ".",
     "encoders/builtin",
     "util",
   ]
-  pruneopts = ""
-  revision = "fb0396ee0bdb8018b0fef30d6d1de798ce99cd05"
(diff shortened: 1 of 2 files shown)


deadprogram:
I was not able to merge this specific PR due to merge conflict with some other commits that I made. But I used the fix suggested here in this commit https://github.com/hybridgroup/gobot/commit/5710da3d096611d5cee0f479e4bc9b419062c247

Thank you @dlisin now closing.

PR #662, opened 2019-04-09
fix bug.

jyk1987:


gen2thomas:
The same thing is fixed with this commit:
https://github.com/hybridgroup/gobot/commit/b3574b310f8b88d9796fb3f2c1aa6137cd02d9bd

Issue #664, opened 2019-04-18
APA102 LEDs driver: bad settings for brightness


air1bzz:
According to this [official documentation](https://cdn-shop.adafruit.com/datasheets/APA102.pdf), the LED Frame 32 bits starts with a 3 bits sequence (111) and 5 bits (Global) for brightness. The brightness cannot exceeds 31.
In the actual code the RGBA Red byte is set instead of Global making unexpected behaviour to the LED strip.

air1bzz:
Above is my fix for this driver, i'm running it successfully on a project.

deadprogram:
This code was release in v1.13.0 thank you! Now closing.

PR #666, opened 2019-05-09
Fix brightness apa102

air1bzz:
#664

### DIFF
--- a/drivers/spi/apa102.go
+++ b/drivers/spi/apa102.go
@@ -6,7 +6,7 @@ import (
 	"gobot.io/x/gobot"
 )
 
-// APA102Driver is a driver for the APA102 programmable RGB LEDs
+// APA102Driver is a driver for the APA102 programmable RGB LEDs.
 type APA102Driver struct {
 	name       string
 	connector  Connector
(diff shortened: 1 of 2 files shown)


deadprogram:
Hi @air1bzz thank you very much for the PR.

I think that it would probably be better to use the Alpha channel as the brightness, instead of having the additional wrapper class for `Color`. That seems to fit the spirit of the API.

What do you think?

air1bzz:
Yes, you're right, I initially wanted to use `Color` in a constructor to avoid brigthness overbounds (0-31). But I found it quite verbose and @MikeZange's proposal in #667 is definitely what we need.

I would like to keep the idea of setting a default brightness in driver's initialization:
if you want to control brightness only with Alpha from `color.RGBA` you can set driver's brightness to `0`.

deadprogram:
Seems like a good idea.

deadprogram:
Hi @air1bzz I merged this commit into `dev`  https://github.com/hybridgroup/gobot/commit/70fd202030bfa3fa0898c837bce460f9a702ecf5 so you should probably close this PR and open a new one with whatever other changes you propose.

Thank you!

air1bzz:
Closing PR, new one updated at https://github.com/hybridgroup/gobot/pull/671

PR #667, opened 2019-05-10
Stop using Red parameter for brightness value

MikeZange:
This change corrects using Red parameter of RGBA for the Brightness value, to using the Alpha parameter of RGBA.

As the brightness max value for APA102 is `31` the `math.Min` stops the driver from using any value higher than 31.

### DIFF
--- a/drivers/spi/apa102.go
+++ b/drivers/spi/apa102.go
@@ -1,9 +1,9 @@
 package spi
 
 import (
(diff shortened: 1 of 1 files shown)


deadprogram:
Hi @MikeZange I cherry picked the commit from this branch into the `dev` branch, so now closing this PR. Thank for for the contribution!

PR #668, opened 2019-05-11
Staticcheck lint fixes

delaneyj:
When creating some new drivers noticed a host of warning from staticcheck/govet/golint.  Was able to squash all but a single unsafe pointer warning that would have to investigate further.

### DIFF
--- a/drivers/i2c/adafruit_driver.go
+++ b/drivers/i2c/adafruit_driver.go
@@ -327,7 +327,7 @@ func (a *AdafruitMotorHatDriver) setPin(conn Connection, pin byte, value int32) 
 	if value == 1 {
 		return a.setPWM(conn, pin, 4096, 0)
 	}
-	return errors.New("Invalid pin")
+	return errors.New("invalid pin")
 }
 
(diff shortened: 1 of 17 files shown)


gen2thomas:
Hello @delaneyj , thank you for your time to fix such issues. 

Unfortunately this changes cause some test errors in CI build. Please could you fix it?

"go fmt" should create the same results and will not result in continuous back and forth but for the linting tool this could be another story. I personally use [golint](https://github.com/golang/lint), which is [deprecated and frozen](https://github.com/golang/go/issues/38968) since 2020. Also gobot mentioned it to [use before contributing](https://github.com/hybridgroup/gobot/blob/d8081f912d324f570c6e3fa74aba7c203e53bb27/CONTRIBUTING.md?plain=1#L38).

Please could you provide some information about your used linting tool.

delaneyj:
It was 3 years ago.  I moved to using Zig+C drivers

gen2thomas:
Hi @delaneyj thank for finding the time to answer and close this PR.

Issue #648, opened 2018-12-31
NewRelayDriver should allow for creating inverted relay


andrewebdev:
Currently `.On()` sets the relay state to high, and `.Off()` sets the relay state to low.

We have realays boards in one of our projects that are actually inverted.
This causes us to do things in the code like `relayDriver.On()` when we want it switched off, and of course it leads to some confusion down the line.

It may be a good idea for NewRelayDriver to allow for this use case?

deadprogram:
Seems like a good idea. A PR that added this would be gladly accepted.

andrewebdev:
@deadprogram just a ping to remind that I have done a PR for this. (just in case this issue was overlooked)

kneerunjun:
```go
func (l *RelayDriver) On() (err error) {
	if err = l.connection.DigitalWrite(l.Pin(), 1); err != nil {
		return
	}

	if l.Inverted {
		l.high = false
	} else {
		l.high = true
	}

	return
}
```
I notice you are still writing 1 to DigitalWriter (Outside the Inverted if logic) in Driver.On - this is hardly helping. Most of the relays close NO-C on LOW ! Wouldn't it help to check Inverted before you DigitalWrite 1.

stuartleeks:
As @kneerunjun noted, `On()` always writes `1` and `Off()` always write `0`. I expected `inverted` to invert the value that was written as well. Unfortunately this means that I'm still having to use `Off()` to turn on the component attached to the relay.

Would you accept a PR that takes the `Inverted` field into account when writing?

deadprogram:
Yes, please, and thanks!

stuartleeks:
Great :-)

andrewebdev:
@stuartleeks Thank for this. I saw this email yesterday and was going to start updating this today; Then I noticed you already submitted a pull request :)

Not sure how I missed that it still writes 1 instead of 0. Our code that uses this functionality is currently still doing the old manual if-check, so we never actually caught this bug in our code.

Good job catching it!

stuartleeks:
Thanks @andrewebdev. I was working on a home project with my raspberry pi and thought I'd misunderstood gobot or mis-wired my relay. After checking my circuits and code I started digging in to the gobot code and found what @kneerunjun described which led me to this issue. 

Hopefully we can get the PR merged and the change released soon, and then I can use the `Inverted` setting and remove the big `NOTE` comment I have in my code explaining the use of `Off()` for `On()` :grin:

[...]

PR #674, opened 2019-06-02
Added ability to make a relay driver inverted

andrewebdev:
I did not want to break any existing projects that already use the RelayDriver, but at the same time didn't want to create a completely new RelayDriver that has all the same functionality as the standard one, execpt that it's inverted.

For this reason I didn't updated the signature for `NewRelayDriver` and rather opted to add `SetInverted` which can then set this value after the initial instance was created.

[#648]

### DIFF
--- a/drivers/gpio/relay_driver.go
+++ b/drivers/gpio/relay_driver.go
@@ -8,7 +8,7 @@ type RelayDriver struct {
 	name       string
 	connection DigitalWriter
 	high       bool
-	inverted   bool
+	Inverted   bool
 	gobot.Commander
 }
 
@@ -24,7 +24,7 @@ func NewRelayDriver(a DigitalWriter, pin string) *RelayDriver {
 		pin:        pin,
 		connection: a,
 		high:       false,
-		inverted:   false,
+		Inverted:   false,
 		Commander:  gobot.NewCommander(),
 	}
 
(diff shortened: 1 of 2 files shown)


drivers/gpio/relay_driver.go
```diff
@@ -23,6 +24,7 @@ func NewRelayDriver(a DigitalWriter, pin string) *RelayDriver {
 		pin:        pin,
 		connection: a,
 		high:       false,
+		inverted:   false,
```
deadprogram: Seems like just making this public e.g. `Inverted` would let you get rid of the get/set functions below, since they do not do anything other than allowing the value to be changed.

What do you think?

drivers/gpio/relay_driver.go
```diff
@@ -53,6 +55,12 @@ func (l *RelayDriver) Name() string { return l.name }
 // SetName sets the RelayDrivers name
 func (l *RelayDriver) SetName(n string) { l.name = n }
 
+// Inverted returns the RelayDrivers inverted state
```
deadprogram: See comment above.

drivers/gpio/relay_driver.go
andrewebdev: Yes I can do that and agree it would be better. I just tried to follow the convention used for the other fields.

andrewebdev:
@deadprogram updated the PR; `Inverted` is now public

deadprogram:
Thanks for working on this and making the requested changes, @andrewebdev now merging.

PR #678, opened 2019-06-11
Fix typo

kvartborg:
Hey, thanks for the gobot project!

I just spottet a typo, here is a correction.


deadprogram:
Oops, thanks for the correction @kvartborg now merging.

Issue #682, opened 2019-07-13
Should mqtt Adaptor provide access to underline paho.Client instance?


andrewebdev:
So I've seen A couple of instances where people are asking for functionality in `mqtt.Adaptor` that the `paho.Client` already provides.

I suspect gobot it attempting to wrap some of the tedious interfaces to make it easier to interface with the `paho.Client`. But I'm also in the case at the moment where I need to extra functionality that the `paho.Client` already provides ( `IsConnected()`  in my case). It feels a bit cumbersome to have to create wrappers for every method or field that we may want to have access to.

Can we possibly make the `mqtt.Adaptor.client` field public?

This would give us the opportunity to drop down to a lower level if we need to go deeper into the internals, rather than having to create pull requests just for a simple wrapper.

PS: I can contribute a PR to this if this is something useful and if gobot team agrees.

deadprogram:
I think this is an excellent idea. Yes, please.

Issue #445, opened 2017-08-11
Support for "retain" in MQTT messages


Remo:
I'm trying to send a message to an MQTT broker, but retain it. I did some digging and it seems like paho already supports this. The third parameters sets it to `false`: https://github.com/hybridgroup/gobot/blob/master/platforms/mqtt/mqtt_adaptor.go#L124

I'm new to go and gobot and thus wasn't able to create a pull request. Not sure how I'm uspposed to add this without optional arguments or method overloading in a way that doesn't break backwards compatibility.

Also, why not return the result of `a.client.Publish`  so that we can track the delivery of the message?

deadprogram:
Hi @Remo thanks for the comments. Seems like a good idea to implement.

Regarding optional arguments, the best way to handle this is using variadic configuration ala https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis

Gobot's i2c drivers all already do this, to provide a slightly more in-context example: https://github.com/hybridgroup/gobot/blob/master/drivers/i2c/bme280_driver.go#L42

Returning a paho.Token from Publish() certainly seems like a small but useful change, as well.

We'd be very happy to accept a PR with these changes. Please let me know if I can help out in any way.

Remo:
@deadprogram thanks! I'm still evaluating, not sure I'll spend time learning go and gobot at this point. Currently playing around with python libraries and framework.

andrewebdev:
@Remo I've just submitted a PR that should allow you to access `paho.Client` directly from the gobot `mqtt.Adaptor`.

See andrewebdev:682-public-paho-client

gen2thomas:
@deadprogram it seems the originated problem is already fixed with the merged PR #719 . Maybe you could close the issue?

PR #683, opened 2019-07-19
[#682] Make the paho.Client on mqtt.Adaptor public

andrewebdev:
This will allow the access to the underlying `paho.Client` so that we can use it's methods and properties directly, without needing to create new wrappers and methods to access said methods.

### DIFF
--- a/platforms/mqtt/mqtt_adaptor.go
+++ b/platforms/mqtt/mqtt_adaptor.go
@@ -33,7 +33,7 @@ type Adaptor struct {
 	clientKey     string
 	autoReconnect bool
 	cleanSession  bool
-	client        paho.Client
+	Client        paho.Client
 	qos           int
 }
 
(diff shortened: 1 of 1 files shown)


platforms/mqtt/mqtt_adaptor.go
```diff
@@ -33,7 +33,7 @@ type Adaptor struct {
 	clientKey     string
 	autoReconnect bool
 	cleanSession  bool
-	client        paho.Client
+	Client        paho.Client
```
oleg-balunenko: maybe it is better to create a getter instead?

deadprogram: That is probably a good idea. We likely do not want to replace the entire client, just be able to call into it.

gen2thomas:
I will going to close this PR because:
* making the client exported seems to be not accepted
* there is no progress on this PR since 3 years (e.g. with providing a implementation for getter function)
* the suggested implementation in the related issue #445  is not used

Issue #689, opened 2019-09-12
Error installing while using Go Modules due to outdated dependency name


jshields:
When installing using the instructions from https://gobot.io/ using Go 1.13 and Go Modules, an error occurs like so
```
~/workspace/dronez(master): go get -u gobot.io/x/gobot/...
...
go: github.com/codegangsta/cli@v1.22.1: parsing go.mod: unexpected module path "github.com/urfave/cli"
go: error loading module requirements
```

This is due to the usage of the dependency "github.com/codegangsta/cli", which has been renamed to "github.com/urfave/cli"

(See that https://github.com/codegangsta/cli redirects to https://github.com/urfave/cli)

https://github.com/hybridgroup/gobot/blob/635adea96f788ab19bf77470b075f8db44d34fa0/cli/main.go#L6

deadprogram:
Got it. Would you like to submit a PR that corrects this?

jshields:
@deadprogram Sure thing :) Gotta install dep to update the lock file, then I'll send that over.

jshields:
I opened https://github.com/hybridgroup/gobot/pull/690
Feel free to suggest another version of `dep` I should use or different settings for generating the lock file.

MShoaei:
This issue is kind of forgotten but maybe it's a good time to consider migrating to go modules. Also there are **2 dependencies** that need to be fixed and the second one is in platform/nats. I have them fixed and also tried migrating to go modules and I'm  ready to make the PR.
Just making sure if you want to consider migrating to go modules. Here's my [repo](https://github.com/MShoaei/gobot) if you want to check it out.

jshields:
@MShoaei Agreed, moving to Go Modules is the long term solution.

deadprogram:
I have just merged the `dep` PR, and would gladly accept a PR to move to modules!

deadprogram:
The modules code is now in the `dev`` branch.

deadprogram:
This has now been released. Thanks everyone!

PR #690, opened 2019-09-13
689 update codegangsta to urfave

jshields:
This PR seeks to address installation issues stemming from a Git repo that was renamed:
https://github.com/hybridgroup/gobot/issues/689

The only remaining reference to the old name should be in this comment:
https://github.com/jshields/gobot/blob/master/api/basic_auth.go#L11

To update `Gopkg.lock`, I ran these commands:
```
brew install dep
dep ensure
```
I have no insight into why dep made all the changes it did to the lock file, aside from removing the old dependency name and adding the new one. Here is the version I am using:
```
dep version
dep:
 version     : v0.5.4
 build date  : 2019-06-14
 git hash    : 1f7c19e
 go version  : go1.12.6
 go compiler : gc
 platform    : darwin/amd64
 features    : ImportDuringSolve=false
```


MShoaei:
Why isn't this being merged?

deadprogram:
We need to switch to modules. In the meantime, merging this PR. Thank you for the contribution!

PR #697, opened 2019-09-24
API package docs' example fix

ksputo:
Link in the commented section of the example code was not valid.


deadprogram:
Thank you for the documentation fix @ksputo now merging.

PR #701, opened 2019-10-25
Added RPi4 to documentation

conejoninja:
Didn't test it myself, but the GPIO should be compatible with the new model


deadprogram:
Seems legit :+1: merging.

PR #707, opened 2019-11-26
add ultrasonic support for i2c driver

merlindorin:
Add new i2c feature for grovepi. I use this code with the grove ultra sonic ranger : https://www.seeedstudio.com/Grove-Ultrasonic-Distance-Sensor.html

Feel free to comment, accept or refused ^^

btw : nice project ;)


camdenorrb:
Why was this closed?

PR #713, opened 2019-12-23
added support for EDU drones

spacemandev-git:
Tello EDU drones have AP mode which allows a tello to connect to non standard IP address. The default driver function currently has the IP addresses hard coded, not allowing for the AP mode functionality. 

I added a new function, NewEDUDriver which takes in the address to set for the driver.


spacemandev-git:
will branch this and redo the pull request

PR #714, opened 2019-12-23
added newedudriver function

spacemandev-git:
Tello EDU drones have the ability to be put into AP mode, connecting with a router for custom IP addresses. The current Tello driver hard codes the default IP address, which disallows this functionality. The new EDU driver keeps the existing driver while also adding a second driver which can be used for EDU drones with custom IP addresses

### DIFF
--- a/platforms/dji/tello/driver.go
+++ b/platforms/dji/tello/driver.go
@@ -221,6 +221,33 @@ func NewDriver(port string) *Driver {
 	return d
 }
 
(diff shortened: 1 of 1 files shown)


deadprogram:
Hi @Brownie79 I think that this PR I just merged does pretty much the same thing as your PR: https://github.com/hybridgroup/gobot/pull/718

Now closing, but thank you for the contribution.

PR #718, opened 2020-01-04
Added option to specify IP for Tello edu edition

danacr:
Since the Tello edition has the capability to join wifi networks, it gets assigned a different IP address than the default one.

Example implementation: https://github.com/danacr/drone-pod/blob/master/main.go

### DIFF
--- a/platforms/dji/tello/driver.go
+++ b/platforms/dji/tello/driver.go
@@ -221,6 +221,34 @@ func NewDriver(port string) *Driver {
 	return d
 }
 
+// NewDriverWithIP creates a driver for the Tello EDU drone. Pass in the ip address and UDP port to use for the responses
+// from the drone.
+func NewDriverWithIP(ip string, port string) *Driver {
+	d := &Driver{name: gobot.DefaultName("Tello"),
+		reqAddr:   ip + ":8889",
+		respPort:  port,
+		videoPort: "11111",
+		Eventer:   gobot.NewEventer(),
+	}
+
+	d.AddEvent(ConnectedEvent)
+	d.AddEvent(FlightDataEvent)
+	d.AddEvent(TakeoffEvent)
+	d.AddEvent(LandingEvent)
+	d.AddEvent(PalmLandingEvent)
+	d.AddEvent(BounceEvent)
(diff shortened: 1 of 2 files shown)


danacr:
CI fails, but it is unrelated to the code change

trevrosen:
@danacr thanks for the PR, and apologies for the late response!

@deadprogram I'm happy to land this if we can make sure Travis is pleased.

danacr:
Fixed the CI by updating the `.travis.yml` file and added a temporary fix for https://github.com/bugst/go-serial/issues/71

platforms/dji/tello/driver.go
```diff
@@ -195,9 +195,9 @@ type Driver struct {
 
 // NewDriver creates a driver for the Tello drone. Pass in the UDP port to use for the responses
 // from the drone.
-func NewDriver(port string) *Driver {
+func NewDriver(ip string, port string) *Driver {
```
deadprogram: The only issue here is that this is a breaking API change.

Perhaps a better way to achieve this is to add another function `NewDriverWithIP(ip string, port string)` so that existing code continues to work, but that this functionality can be provided.

What do you think?

danacr: I was worried about the same thing, but wasn't certain about a way to do it. I like the idea, let me have a go at it

deadprogram:
I just switched this PR to the `dev` branch as mentioned in our contributions guidelines here:
https://github.com/hybridgroup/gobot/blob/master/CONTRIBUTING.md#general-guidelines

I made a comment inline in the code with a suggestion of adding a new function instead of breaking the current API for other users..

Also perhaps you would like to submit the fixes for Travis as a separate PR? We need that very much right away, thank you so much for reaching/fixing that.

Thank you!

danacr:
Ok, splitting into a separate PR: https://github.com/hybridgroup/gobot/pull/720

platforms/dji/tello/driver.go
danacr: fixed

deadprogram:
Thank you for this contribution @danacr now merging.

Issue #700, opened 2019-10-20
Spi panics when closing program with Raspberry and mcp3008


tux-eithel:
Hello!

I'm using gobot @ 1.14.0, but the problem occurs also with 1.13.0.

I'm having a panic when I hit ctrl-c to close my program. Let's see the details.

## What 

I'm doing a simple example with mcp3008 chip and RaspberryPI. All works fine, until ctr-c is hit and the software crash. This is the stack:

```
^C2019/10/20 14:57:58 Stopping Robot Light Sensor ...
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0xec6a4]

goroutine 1 [running]:
periph.io/x/periph/host/sysfs.(*SPI).Close(0x1cb2000, 0x0, 0x0)
	/home/example-user/Go/pkg/mod/periph.io/x/periph@v3.6.2+incompatible/host/sysfs/spi.go:62 +0x70
gobot.io/x/gobot/drivers/spi.(*SpiConnection).Close(0x1c840c0, 0x1c6a8c0, 0x1c63e1c)
	/home/example-user/Go/pkg/mod/gobot.io/x/gobot@v1.14.0/drivers/spi/spi.go:69 +0x28
gobot.io/x/gobot/platforms/raspi.(*Adaptor).Finalize(0x1c90120, 0x0, 0x0)
	/home/example-user/Go/pkg/mod/gobot.io/x/gobot@v1.14.0/platforms/raspi/raspi_adaptor.go:123 +0x2fc
gobot.io/x/gobot.(*Connections).Finalize(0x1c740b0, 0x0, 0x0)
	/home/example-user/Go/pkg/mod/gobot.io/x/gobot@v1.14.0/connection.go:64 +0x78
gobot.io/x/gobot.(*Robot).Stop(0x1c86050, 0x0, 0x1c82240)
	/home/example-user/Go/pkg/mod/gobot.io/x/gobot@v1.14.0/robot.go:208 +0xb8
gobot.io/x/gobot.(*Robot).Start(0x1c86050, 0x0, 0x0, 0x0, 0x1, 0x1c84060)
	/home/example-user/Go/pkg/mod/gobot.io/x/gobot@v1.14.0/robot.go:194 +0x1f0
main.main()
	/home/example-user/Go/src/gitlab.com/eithel/stupid-smart-things/mcp/main.go:38 +0x1bc
```

[...]

PR #722, opened 2020-01-13
#700 - Avoid to close the connection.

tux-eithel:
Each spi drivers must not close the connection: it's not its responsibility.
Robots should close the connection.

Signed-off-by: Eithel <tux.eithel@gmail.com>


deadprogram:
Thank you for contributing the fix for this @tux-eithel 

The `Adaptor` is what closes the connection to be more precise.

In any case, thanks again. Now merging.

PR #728, opened 2020-02-14
Add INPUT_PULLUP support on firmdata platform

disaster37:
Add the availability to use INPUT_PULLUP on some GPIO driver:
  - button
  - direct_pin
  - makey_button
  - pir_motion_driver

INPUT_PULLUP on arduino is use to avoid to add some resistor when yous should to read the state of button.
When INPUT_PULLUP is enable, the signal is inverted, so the platform invert signal to be transparent for user.

Sample to use it with button driver:
```go
func main() {
        firmataAdaptor := firmata.NewAdaptor("/dev/ttyACM0")
        button := gpio.NewButtonDriverDriver(firmataAdaptor, "13")
        err := button.SetInputPullup()
        if err != nil {
          panic("Not supported by plateforme")
        }
        work := func() {
                buton.On(gpio.ButtonPush, func(data interface{}) {
                  fmt.Printf("Button pushed")
                }
        }

        robot := gobot.NewRobot("bot",
                []gobot.Connection{firmataAdaptor},
                []gobot.Device{button},
                work,
        )

        robot.Start()
}
```

> Button must be plug on input pin and in GND instead of 5v

### DIFF
--- a/drivers/gpio/aip1640_driver.go
+++ b/drivers/gpio/aip1640_driver.go
@@ -21,7 +21,7 @@ const (
 //
 // Library ported from: https://github.com/wemos/WEMOS_Matrix_LED_Shield_Arduino_Library
 type AIP1640Driver struct {
-	pinClock  *DirectPinDriver
+	pinClock   *DirectPinDriver
 	pinData    *DirectPinDriver
 	name       string
 	intensity  byte
@@ -34,7 +34,7 @@ type AIP1640Driver struct {
 func NewAIP1640Driver(a gobot.Connection, clockPin string, dataPin string) *AIP1640Driver {
 	t := &AIP1640Driver{
 		name:       gobot.DefaultName("AIP1640Driver"),
-		pinClock:  NewDirectPinDriver(a, clockPin),
+		pinClock:   NewDirectPinDriver(a, clockPin),
 		pinData:    NewDirectPinDriver(a, dataPin),
 		intensity:  7,
 		connection: a,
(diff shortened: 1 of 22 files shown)


disaster37:
Hi,

there are some reason to not merge this PR ?

gen2thomas:
@disaster37 thank you for your PR. IMO this is a nice new feature, so I will try to make some suggestions to get some progress.
* please resolve conflicts by rebase to the current state of dev
* please revert all changes made by "go fmt" in unrelated packages, because this leads to harder review
* please revert the change in go.sum, if this is not important for the work of your new code
* please provide a working example in "gobot/examples" instead of (or in addition to) the sample in your description, this will help future users

Maybe I have not understand the code quite correctly, but would it be better to add an option for each input instead of introducing a new implementation for DigitalRead(), e.g. by an interface with "SetPullUp(pin string, state bool)" and "IsInputPullup()"? In general this should be possible, because the "f.Board.Pins()[p].Mode" is used in both "DigitalRead()" and "DigitalReadInputPullup()".

Instead of "go fmt" in unrelated packages and adjust/cleanup of  "go.mod/go.sum" you can add an issue (maybe there is one for "go fmt") or try to solve such general format errors by a separate PR.

Although I have no firmata adaptor tested yet, thanks again for this new feature! IMO this could be also useful for other boards.

gen2thomas:
I will now revert the PR back to Draft as reference

gen2thomas:

Issue #564, opened 2018-06-10
Joystick needs a way to handle multiple joysticks


deadprogram:
Right now, the code only connects to the first joystick i.e. joystick 0:

https://github.com/hybridgroup/gobot/blob/master/platforms/joystick/joystick_adaptor.go#L30

We need the ability to specify a particular joystick to handle case that more than one joystick is connected ta the same time.

carl-ranson:
I did some tinkering on this today and managed to come up with the following joystick adaptor newAdaptor function. You pass in the joystick name like NewAdaptor("Magicsee R1"). 

hope that helps with resolving this ticket. 

func NewAdaptor(targetName string) *Adaptor {
	return &Adaptor{
		name: gobot.DefaultName("Joystick"),
		connect: func(j *Adaptor) (err error) {
			sdl.Init(sdl.INIT_JOYSTICK)
			foundIndex := -1
			for i := 0; i<sdl.NumJoysticks(); i++ {
				if strings.TrimSpace(targetName) == strings.TrimSpace(sdl.JoystickNameForIndex(i)) {
					foundIndex = i
				}
			}
			if foundIndex > -1 {
				j.joystick = sdl.JoystickOpen(foundIndex)
				return
			}
			return errors.New("No joystick available")
		},
	}
}

theunkn0wn1:
Greetings, is there any progress on this? 

I have a project that needs to be able to handle multiple joysticks, and this only appears to work when @carl-ranson 's patch is applied.

gen2thomas:
there are two implementations available for reference:
* #715 
* #732

gen2thomas:
To earn the advantages of both implementations we could do the following:
* use a string for the identifier
* try to parse the number from string - if ok, try to use it -> (*)
* otherwise try to find the number by name (use first finding)
* optional: log a warning or return an error, if there are more than one devices with the same name

(*) If this fails, maybe the name is a number? If this can occur, try to find by name.

PR #732, opened 2020-03-07
Added ability to select a joystick by name

theunkn0wn1:
Tihs PR adds a new Adaptor constructor for creating a joystick by implementation-specific name.

relates to #564

### DIFF
--- a/platforms/joystick/joystick_adaptor.go
+++ b/platforms/joystick/joystick_adaptor.go
@@ -2,6 +2,7 @@ package joystick
 
 import (
 	"errors"
+	"strings"
 
 	"gobot.io/x/gobot"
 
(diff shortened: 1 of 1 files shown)


deadprogram:
Hello @theunkn0wn1 thank you for the contribution.

How does the naming work if you have 2 of the same joysticks connected?

Could you please add an example of how you would use this with 2 joysticks? Ideally this would be in the godocs comments for the new constructor.

theunkn0wn1:
To my understanding, this implementation has the same inherent flaw as the constructor it replaces, this implementation will always return the first of the two same-model joysticks as the underlying SDL2 library reports them. As I don't have two of the same joystick to compare against, im not sure what other identifiers on a joystick may be more unique we could base the constructor on.

I will happily provide godoc formatted examples, though I appear to be having a hard time finding the relevant documentation. :man_facepalming: Can you provide me a link for working with godoc?

gen2thomas:
@theunkn0wn1 because there is no answer yet by @deadprogram, I will try to give some help. 

IMO there is no specific format for the first raw documentation needed. Just have a look at some other documentation of constructors in gobot. If there any developer/user is unsure about the correct use he will have a look into some test code or even an example (e.g. gobot/examples/*). Unit tests in general are very helpful.

theunkn0wn1:
thanks @gen2thomas for a reply. The lack of activity on this PR caused me to completely forget it hadn't been seen to.
I have long since moved on from golang in general, so this PR is now offered as-is.

gen2thomas:
@theunkn0wn1 thanks for the reply. So I will set it to draft as a reference.

gen2thomas:

PR #739, opened 2020-03-20
Support for SHT2x Temperature/Humidity sensor driver

hnw:
New driver for Sensirion SHT2x (SHT20/SHT21/SHT25).

I tested this on Raspberry Pi Zero with SHT21.


deadprogram:
Thanks for this contribution @hnw now merging.

PR #742, opened 2020-03-31
Update value written via DigitalWrite if RelayDriver.Inverted is true

### LINKED ISSUE #648: NewRelayDriver should allow for creating inverted relay

stuartleeks:
The goal of this PR is to update the RelayDriver to write `0` instead of `1` for `On()` if `Inverted` has been set `true` (and similar inversion for `Off()`, as per [this comment](https://github.com/hybridgroup/gobot/issues/648#issuecomment-605701451) and [initial agreement](https://github.com/hybridgroup/gobot/issues/648#issuecomment-605701691) from @deadprogram 

I ended up making more changes in this PR than I initially expected, but they are broken into three separate commits if that helps:

* Update `gpioTestAdapter` so that the signatures of functions to provide behaviour match the `DigitalWriter` interface methods etc. This paves the way for capturing the values that are written
* Update the `RelayDriver` tests to capture and validate the values that are written via `DigitalWrite`
* Update the RelayDriver to write `0` instead of `1` for `On()` if `Inverted` has been set `true` (and similar inversion for `Off()`

Apologies for the extra changes - I wanted to ensure that there were tests for this change. If there's an alternate way to perform these tests that you'd prefer then let me know.

I believe that this closes #648

### DIFF
--- a/drivers/gpio/button_driver_test.go
+++ b/drivers/gpio/button_driver_test.go
@@ -44,7 +44,7 @@ func TestButtonDriverStart(t *testing.T) {
 		sem <- true
 	})
 
-	a.TestAdaptorDigitalRead(func() (val int, err error) {
+	a.TestAdaptorDigitalRead(func(string) (val int, err error) {
 		val = 1
 		return
 	})
@@ -62,7 +62,7 @@ func TestButtonDriverStart(t *testing.T) {
 		sem <- true
 	})
 
-	a.TestAdaptorDigitalRead(func() (val int, err error) {
+	a.TestAdaptorDigitalRead(func(string) (val int, err error) {
 		val = 0
 		return
 	})
@@ -77,7 +77,7 @@ func TestButtonDriverStart(t *testing.T) {
 		sem <- true
 	})
 
-	a.TestAdaptorDigitalRead(func() (val int, err error) {
+	a.TestAdaptorDigitalRead(func(string) (val int, err error) {
 		err = errors.New("digital read error")
 		return
 	})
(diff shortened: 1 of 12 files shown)


stuartleeks:
Just wondered if there were any comments on this?

trevrosen:
@deadprogram I don't have an issue w/ merging this. Lemme know how you feel about the alterations to RelayDriver.

deadprogram:
First of all, thanks for the contribution.

I think I am having trouble digesting the changes to the tests. Maybe that should have been in a separate PR?

In any case, the changes to `RelayDriver` are quite minimal and make sense.

@trevrosen if you are OK with this, then I am as well.

stuartleeks:
@deadprogram yeah, sorry! As I mentioned in the initial comment, I split the PR into commits for each area in the hope that that made it easier to see the different sets of changes. I'm happy to split out the `gpioTestAdapter` changes from the `RelayDriver` changes if you prefer?

stuartleeks:
@deadprogram - I have created #745 which just contains the change for the tests and added some more description in the PR notes. I hope that helps to make the test changes clearer (and am happy to rebase this PR onto #745 if you prefer to do that).

Shout if you have any questions :-)

stuartleeks:
@deadprogram - qq, when is the next release likely to be? (i.e. when will this change be included in a release)

PR #745, opened 2020-04-05
Allow capture of values written to gpioTestAdapter for asserting in tests

stuartleeks:
As I started working on the changes for #742 I wanted to be able to verify the values written via `DigitalWrite` in the unit tests for `RelayDriver`.

To do this I had to align the function signatures in `gpioTestAdaptor`.

For example, this is the `DigitalWrite` signature:

```go
func (t *gpioTestAdaptor) DigitalWrite(pin string, val byte) (err error) 
```

But the function that can be injected to control the behaviour of the test adapter has this signature:

```go
func() (err error) 
```

This core of this PR is to change the signature of the injected function to align with the `DigitalWrite` Signature:

```go
func(pin string, val byte) (err error)
```

The same change has been made for the other functions in `gpioTestAdapter`.

Most of the rest of the changes in the PR are updating code so that it still compiles with these changes, but there are a couple of tests for `RelayDriver` that have been updated to supply a function that captures the value written via `DigitalWrite` so that the test can assert that it is the expected value.

### DIFF
--- a/drivers/gpio/button_driver_test.go
+++ b/drivers/gpio/button_driver_test.go
@@ -44,7 +44,7 @@ func TestButtonDriverStart(t *testing.T) {
 		sem <- true
 	})
 
-	a.TestAdaptorDigitalRead(func() (val int, err error) {
+	a.TestAdaptorDigitalRead(func(string) (val int, err error) {
 		val = 1
 		return
 	})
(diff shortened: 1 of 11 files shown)


trevrosen:
@stuartleeks I merged #742 already so I assume this one is now redundant? Thanks for your patch and your interest in Gobot! Feel free to close this if no further action is needed.

stuartleeks:
Great, I'll close this PR as it is not needed. Thanks

PR #752, opened 2020-05-03
Use go.mod-friendly version of go.bug.st/serial package

dlisin:
Replace `go.bug.st/serial.v1` with go.mod-friendly `go.bug.st/serial`


deadprogram:
Thanks @dlisin for the contribution. Now merging.

PR #757, opened 2020-06-22
Addded xbox one controller to the list

wlwatkins:
Hope this helps, there is only one main modification which is theconfig json where two ids where switched between the 360 and the one

### DIFF
--- a/platforms/joystick/joystick_driver.go
+++ b/platforms/joystick/joystick_driver.go
@@ -25,6 +25,10 @@ const (
 
 	// Xbox360RockBandDrums controller configuration.
 	Xbox360RockBandDrums = "xbox360RockBandDrums"
+	
+	// Xbox360 joystick configuration.
+	xboxONE = "xboxONE"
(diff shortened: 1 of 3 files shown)


platforms/joystick/joystick_driver.go
```diff
@@ -25,6 +25,10 @@ const (
 
 	// Xbox360RockBandDrums controller configuration.
 	Xbox360RockBandDrums = "xbox360RockBandDrums"
+	
+	// Xbox360 joystick configuration.
+	xboxONE = "xboxONE"
```
krobertson, line 30: This needs to be changed to `XboxONE` as the constant name.

deadprogram:
Also, needs `go fmt` run on the changed files, please.

gen2thomas:
done now by #871 , thanks for your work and sorry for the long delay

wlwatkins:
Wow this i had forgotten about ^^

PR #760, opened 2020-08-02
i2c/pca9685: Resolve issue with PWM for PWMWrite

imle:
4096 is outside of the value that is able to be written to a pin. Lowered the scale and updated the comments


deadprogram:
Thanks for the correction @imle now merging.

PR #761, opened 2020-08-10
[FIX] stepperDriver Run not using stepper speed

a1xon:
I'm not sure if anyone using it like this? Like blasting the stepperDriver with full force for loop steps.
At least my stepper is making noises like a badly wounded rabbit and is not moving anywhere.
Using a delay in Run() like in Move() fixed that. I'm pretty sure it was intended in that way.

### DIFF
--- a/drivers/gpio/stepper_driver.go
+++ b/drivers/gpio/stepper_driver.go
@@ -119,12 +119,15 @@ func (s *StepperDriver) Run() (err error) {
 	s.moving = true
 	s.mutex.Unlock()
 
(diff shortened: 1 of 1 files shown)


trevrosen:
@a1xon hey thanks for this PR! I'd like to review (and possibly merge), but per the contributing guide, please open against `dev` first. Thanks for your interest in Gobot!

a1xon:
oh I am sorry! I was doing this via Github. Will open again against dev.

PR #724, opened 2020-01-15
Add delay for Run function of StepperDriver.

tdaira:
Add delay for Run function of StepperDriver and create common sleep function.
Testing on the actual machine has been completed.

Issue: https://github.com/hybridgroup/gobot/issues/712

PR #762, opened 2020-08-10
[FIX] stepperDriver Run not using stepper speed

a1xon:
I'm not sure if anyone using it like this? Like blasting the stepperDriver with full force for loop steps.
At least my stepper is making noises like a badly wounded rabbit and is not moving anywhere.
Using a delay in Run() like in Move() fixed that. I'm pretty sure it was intended in that way.

Change from original PR: Run should react to speed changes.

Original PR was against Master. I'm sorry. - https://github.com/hybridgroup/gobot/pull/761/commits

### DIFF
--- a/drivers/gpio/stepper_driver.go
+++ b/drivers/gpio/stepper_driver.go
@@ -120,11 +120,14 @@ func (s *StepperDriver) Run() (err error) {
 	s.mutex.Unlock()
 
 	go func() {
+		var delay time.Duration;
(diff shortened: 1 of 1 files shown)


deadprogram:
Hello @a1xon I just merged a different PR that addressed this issue in a similar way. Please see #724

gen2thomas:
@deadprogram the related bug is already merged, can we close this Issue?

PR #768, opened 2020-09-26
Support for BMP388 Barometric Pressure/Temperature/Altitude Sensor

ferllings:


deadprogram:
Looks great @ferllings thank you very much for the contribution.

Now merging.

PR #776, opened 2021-01-05
Remove gobuffalo/uuid archived package

ccary64:
when following the direction on gobot.io running the example code, it will fail to run as the modules will pick up the latest UUID package version (looks like mod version is incompatible). This package version has removed UUID V2 for security reasons and this will cause the archived package gobuffalo/uuid to fail. This package does not effect the base package gofrs/uuid so this PR eliminates it.


deadprogram:
Thanks for the fix for this @ccary64 now merging.

wunderkind2k1:
Thanks for fixing and merging!
So when is the next release planned as I have installed *gobot* via **go get** and not via building the source?

gen2thomas:
Hi @wunderkind2k1 , now it is part of release v1.16.0.

PR #785, opened 2021-02-24
Add gpio interrupt handling, with an example using encoders

erh:
Add support sysfs event handlers.
There is an example for how to use for a motor encoder on a raspberry pi.

Also fixed a minor bug where if you're using a standard L298N motor controller, the motor driver Off() function doesn't properly turn the motor off because it only changes the speed, doesn't actually shut the control wires to low.


gen2thomas:
Hi @erh thank you for your work and sorry for the very long delay. In the meantime there has something changed on target branch, among other things, the character device gpio support  was introduced. This includes support for gpio events and pullup/down too. From your point of view, does it still make sense to invest time in this PR and are you able/want to do that?

gen2thomas:

gen2thomas:
Hi @erh , I'm going to close this PR due to inactivity and there seems to be a working solution now with character device gpio.

PR #790, opened 2021-03-27
fix helper_test ReadByteData, ReadWordData to use reg

### LINKED ISSUE #789: i2c helper_test.go ReadByteData() & ReadWordData() miss usage of "reg"
...therefore unit tests can not test:
* right register address was used
* count of writes (or written bytes) are correct

In general before reading any data, the register must be set by a write operation.

This can be performed by preceding:
```go
if err = t.WriteByte(reg); err != nil {
  return
}
```

gen2thomas:
will fix #789 
* fix helper
* fix drv2605l test, see section 8.5.3.5 of data sheet for single-byte-read
* fix bmp388 test, see section 5.2.2 of data sheet for i2c read


gen2thomas:
@deadprogram please have a look at this PR. It is only related to and improve some tests and general testability, so a merge will be safe and could possibly help to write some tests for #810.

deadprogram:
Thanks @gen2thomas switched this to `dev`branch. Now merging, thanks!

PR #800, opened 2021-08-04
fix raspi PWMPin.SetDutyCycle

DiscreteTom:
0 is allowed in pi blaster


deadprogram:
Thanks @DiscreteTom now squashing/merging.

PR #801, opened 2021-08-07
Add Driver Support for Grove DHT11

snmed:
Signed-off-by: snmed <sandro@dallo.biz>

This commit adds support for the Grove DHT11 Humidity & Temperature Sensor.

-  Add Driver GroveDHT11SensorDriver
-  Add Tests for GroveDHT11SensorDriver
-  Add ReadDHT function for GrovePi Driver

### DIFF
--- a/drivers/aio/aio.go
+++ b/drivers/aio/aio.go
@@ -24,9 +24,3 @@ type AnalogReader interface {
 	//gobot.Adaptor
 	AnalogRead(string) (val int, err error)
 }
-
-// DHTReader interface represents an Adaptor which has capabilities to
-// read digital humidity and temperature for the GrovePi DHT sensors.
-type DHTReader interface {
-	ReadDHT(pin string) (temperature, humidity float32, err error)
-}
--- a/drivers/aio/aio.go
+++ b/drivers/aio/aio.go
@@ -24,3 +24,9 @@ type AnalogReader interface {
 	//gobot.Adaptor
 	AnalogRead(string) (val int, err error)
 }
+
+// DHTReader interface represents an Adaptor which has capabilities to
+// read digital humidity and temperature for the GrovePi DHT sensors.
+type DHTReader interface {
+	ReadDHT(pin string) (temperature, humidity float32, err error)
+}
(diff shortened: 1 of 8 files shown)


snmed:
I moved sensor driver into the gpio package because it uses a digital pin and therefore belongs not into the aio package.

gen2thomas:
Hi @snmed , thanks for providing this driver and sorry for the late reply.

* please can you provide an example, so one can see how to setup/use the device with the driver (e.g. as reference "raspi_grove_pi_blink.go")
* ~~the usage of gpio or aio IMO do not depend on pin usage (AnalogRead, AnalogWrite also needs pin), but on the format of return value, so my vote goes for "aio". But maybe I'm wrong for GrovePi device, if there are different types of pins.~~ (was wrong)
* I would chose a more generic approach:
  * implement a general "humidity_temperature_sensor_driver" in "aio", with usage of new interface "HumidityTemperatureReader" with return values of type float64, (2x AnalogReader can not be used because return type int)
  * add comments for the expected units (celsius, rel. humidity in percent) on interface
  * (optional) provide a possibility to return temperature in "F" in generic implementation
  * implement the specific GrovePi variant based on the general driver (in aio or gpio)

gen2thomas:
I take a look into the documentation of DHT11 and DHT22. It uses a one wire interface with serial communication. Gobot itself has no driver for this type of communication yet. It is on the roadmap, possibly it would be a folder [serial](https://github.com/hybridgroup/gobot/blob/release/ROADMAP.md#serial) with "dht11_driver.go".

GrovePi firmware seems to have [a driver implemented](https://github.com/DexterInd/GrovePi/blob/master/Firmware/Source/grovepi/src/DHT.cpp) to read the DHT one-wire-serial protocol and provide it as i2c information. This i2c-interface is now implemented as "i2c.ReadDHT()".

According to the existing "UltrasonicRead()" it seems there is no need for more than the i2c-driver itself in gobot. IMO this will also prevent some confusion how to use the driver correctly.

gen2thomas:
already done with #870

Issue #813, opened 2022-03-21
support for PCF8591 8-bit 4xADC 1xDAC i2c


gen2thomas:
Hello, the PCF8591 seems to be on the [roadmap](https://github.com/hybridgroup/gobot/blob/release/ROADMAP.md#i2c). Has the work already been started? If yes, how can I support it, if no I would start with the work on a new feature branch.

Thanks Thomas

gen2thomas:
I am too impatient and therefore have started.

gen2thomas:
@deadprogram please have a look at the PR #814 and possibly close the related roadmap item

gen2thomas:
When implementing the driver I noticed the Tinkerboard is running in 400kbit mode by default and the i2c clock not slows down when attaching the PCF8591 device, which only supports 100kbit. This causes some problems with the driver, which can mostly solved with more write/read cycles (a "With" function is added for that). For those who want a most stable application, it is possible to switch to 100kbit mode, at least in Tinkerboard. An additional "clock-frequency = 100000" to the command line does not change anything for my device, so I add an device tree overlay in this way:

create a new overlay text file below "/boot/overlays/" by "sudo nano i2c1-100kbit-tinker-overlay.dts" with the content:
```
// Definitions for i2c1 with 100 kbit
/dts-v1/;
/plugin/;

/ {
	compatible = "rockchip,rk3288-evb-rk808-linux", "rockchip,rk3288";

	fragment@0 {
		target = <&i2c1>;
		__overlay__ {
			status = "okay";
			clock-frequency = <100000>;
		};
	};
};
```
save and close the editor, than run the device tree compiler:
`dtc -@ -I dts -O dtb -o i2c1-100kbit-tinker-overlay.dtbo i2c1-100kbit-tinker-overlay.dts`

after that, add "overlay=i2c1-100kbit-tinker-overlay" (without file extension) to the "/boot/config.txt"  and reboot

PR #814, opened 2022-03-26
Feature/pcf8591

gen2thomas:
this PR will cover the issue #813 
### Adds in detail
* possibility to use AnalogSensorDriver for scaling values to float64
* add possibility to switch off polling with given 0s as polling intervall, although Start() is called
* add generic TemperatureSensorDriver which supports nonlinear NTC, PT100, PTC and other linear sensors
* adjust grove temperatur sensor to the generic one, add tests for correct scaling the NTC with B=3975, R25=10k, Rs=10k
* introduce the AnalogWriter interface
* introduce AnalogActuatorDriver which uses the new interface 
* add driver for PCF8591 to cover #813 
* add driver for YL-40 module which contains the PCF8591, temperature sensor (B=3950, R25=10k, Rs=1k, invers), brightness sensor, potentiometer, free analog input, one analog output, use all previously added features
* examples for PCF8591 and YL-40 module provided
* most features practically tested with tinkerboard

### Notes for NTC scaling
* the scaling can be done by T0, R0, B or also by T0, R0, T1, R1 --> B will be calculated at initialization
* thermistor is typically connected with an series resistor between reference voltage and ground, both directions are supported (normal, inverse)
* all NTC test expectations were calculated by an LibreOffice table

### Notes for the PCF8591 chip and the driver
The IC PCF8591 only supports 100 kHz (basic 100 kbit/s mode), see  [section 14.4 Dynamic characteristics at page 20 of NXP specification](https://www.nxp.com/docs/en/data-sheet/PCF8591.pdf). Running the PCF8591 in 400 kbit fast mode, which seems to be the default of most platforms now, leads to some instability (tested with 5 devices at ~370 kbit/s on tinkerboard):
1. the control byte is not recognized on first write (on 4 devices, ~30% of calls)
2. the read value is to small when immediately read after write (all devices, 100% of calls)
3. the read value is still the old one after more than one reads (all devices, ~99% of calls)
4. the read value is still the old one after more than two reads (4 devices, ~50% of calls) 

strategies to overcome problems with 400 kbit:
for 1.: do "full double reading" (write control register and read, done in most other drivers) fix the problem for 3 tested devices, ~95% good calls
for 3,4.: multiple write don't stabilize this enough (only reduce to ~75%), multiple reading to an buffer and after some relaxation a single read is quite sufficient to reduce it to 5% wrong reads (for one device the problem is gone completely)
for 2.: for long conversion time, write also don't do anything for stabilization, it will be covered by the strategy above

driver:
* all modes are supported (single, 2-/3- differential, mixed) - most other drivers (python, cpp) don't do that
* auto increment is not supported yet
* 400 kbit stabilizing strategies are part of the driver (default is 2x write, 3x buffered read, wait, 1x read)

### Notes for YL-40 module driver
Because the supply voltage of i2c bus on tinkerboard seems to be 3.3V and both LED's are shining very bright, I would suggest to not run this board with 5V. Therefor some scaler are prepared for 3.3V as default. This can be changed by a with-function to each scaler function you want.

### DIFF
--- a/drivers/aio/README.md
+++ b/drivers/aio/README.md
@@ -12,9 +12,11 @@ go get -d -u gobot.io/x/gobot/...
 ## Hardware Support
 Gobot has a extensible system for connecting to hardware devices. The following AIO devices are currently supported:
   - Analog Sensor
+  - Analog Actuator
   - Grove Light Sensor
   - Grove Rotary Dial
   - Grove Sound Sensor
   - Grove Temperature Sensor
+  - Temperature Sensor (supports linear and NTC thermistor in normal and inverse mode)
 
 More drivers are coming soon...
--- a/drivers/aio/aio.go
+++ b/drivers/aio/aio.go
@@ -15,12 +15,20 @@ const (
 	Error = "error"
 	// Data event
 	Data = "data"
+	// Value event
+	Value = "value"
 	// Vibration event
 	Vibration = "vibration"
 )
 
-// AnalogReader interface represents an Adaptor which has Analog capabilities
+// AnalogReader interface represents an Adaptor which has AnalogRead capabilities
 type AnalogReader interface {
 	//gobot.Adaptor
-	AnalogRead(string) (val int, err error)
+	AnalogRead(pin string) (val int, err error)
+}
+
+// AnalogWriter interface represents an Adaptor which has AnalogWrite capabilities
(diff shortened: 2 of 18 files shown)


deadprogram:
Another massive PR, you have me in awe @gen2thomas 

Merging :fireworks:

PR #819, opened 2022-05-01
BUGFIX: new pwm behaviour of tinkerboard

### LINKED ISSUE #818: New ASUS Tinkerboard firmware "TinkerOS" changed pwm path
Currently tinkerboard adaptor uses the sysfs/pwm_pin.go file. The default Path is set to "/sys/class/pwm/pwmchip0" (pwm0 for pin 33 and pwm1 for pin 32). This does not work, at least with the new TinkerOS. After some try and error it seems the path for pin 32 (PWM3) is "/sys/class/pwm/pwmchip3/pwm0" and for pin 33 (PWM2) it is "/sys/class/pwm/pwmchip2/pwm0".

In addition it seems like the call of "newPin.SetPeriod(10000000)" must be done before call of "newPin.InvertPolarity(false)" in the adaptor PWMPin() function.

Also "inverted" is the wrong command for "polarity", it is now "inversed".

This seems to be changed with newer kernels and maybe affect also other adaptors.

I will do some further investigation and can hopefully provide an bugfix for that in a few days.

gen2thomas:
* fix #818 path issue similar to implementation in [beaglebone adaptor](https://github.com/hybridgroup/gobot/blob/release/platforms/beaglebone/beaglebone_adaptor.go). Introduce sysfs.Find() to avoid func in struct.
* fix #818 function order problems in PWMPin().
* fix #818 naming problem with "inverted" by introducing sysfs.SetPolarity()
* read period to use in ServoWrite() - TODO resolved.
* implement SetPeriod()  to ensure automatic adjustment of duty cycle.
* expand error messages in sysfs.pwm_pin.go for better trouble shooting, some other improvements
* adjust tests

### DIFF
--- a/platforms/tinkerboard/README.md
+++ b/platforms/tinkerboard/README.md
@@ -14,15 +14,43 @@ You would normally install Go and Gobot on your workstation. Once installed, cro
 go get -d -u gobot.io/x/gobot/...
 ```
 
-### Enabling GPIO pins
+### System access and configuration basics
 
-To enable use of the Tinker Board GPIO pins, you need to perform the following steps as a one-time configuration. Once your Tinker Board has been configured, you do not need to do so again.
+Some configuration steps are needed to enable drivers and simplify the interaction with your Tinker Board. Once your Tinker Board has been configured, you do not need to do so again.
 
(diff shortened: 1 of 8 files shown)


gen2thomas:
Revert to draft because the name for inverted polarity must be "inversed" instead of "inverse"

gen2thomas:
@deadprogram ready to review & merge

deadprogram:
Thank you very much for working on this @gen2thomas looks good! Now merging.

PR #823, opened 2022-06-06
BUGFIX: Adafruit1109 bad characters on Halt()

### LINKED ISSUE #822: Adafruit1109 driver shows bad characters after Halt()
This happens sometimes and mostly can not be recovered unless power off and on the device.

The root cause is the randomized stop of the used devices MCP23017 and HD44780 in combination with the non existent mutex in the Halt() function of HD44780. This can lead to interruption of any write function of HD44780 display.

The #817 (which was bypassed by #812 ) was possibly an first indicator for such problems.

gen2thomas:
* FIX #822 
* small reordering on initialization procedure according to the documentation
* double the wait time after clear() according to the documentation
* fix some typos, add comments
* mcp23017 now based on i2c.Driver

### DIFF
--- a/drivers/gpio/hd44780_driver.go
+++ b/drivers/gpio/hd44780_driver.go
@@ -54,10 +54,10 @@ const (
 
 // databit pins
 type HD44780DataPin struct {
-	D0 string // not used if 4bit mode
-	D1 string // not used if 4bit mode
-	D2 string // not used if 4bit mode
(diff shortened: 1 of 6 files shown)


gen2thomas:
convert back to draft, because some new ideas for i2c basic reworks

gen2thomas:
going to merge this PR without further review, because:
* changed and tested with Tinkerboard
* unit tests improved

Issue #828, opened 2022-08-30
ads1x15 driver improvements


gen2thomas:
While working on #827 I noted some improvement potential
* different pre-configuration for 4 possible channels with PGA and DR is not possible
* after first configuration for PGA and DR, normally there is no need to change of that for each read (maybe 2 func's can be removed)
* not needed exposed variables and struct items
* wrong/misleading comments
* same code in New... functions
* AnalogRead() is limited to full scale range of 1023, although internally the range is +32767/-32768
* some debug messages would be helpful
* open TODO for With-function
* open TODO for API functions
* WriteBlockData() and ReadBlockData() can be used (after #833 is ready to use)

gen2thomas:
part of release v2.0.0

PR #829, opened 2022-08-30
BUGFIX/IMPROVEMENTS: ADS1x15

### LINKED ISSUE #827: ads1x15 driver not working stable when reading multiple inputs
Currently I'm testing a module with ADS1115 and 2 used analog input pins. Most of the time the values of both inputs are converted as the same, although the voltage differs (measured by a meter).

The main reason seems to be the wait [line 350 in driver code](https://github.com/hybridgroup/gobot/blob/d8081f912d324f570c6e3fa74aba7c203e53bb27/drivers/i2c/ads1x15_driver.go#L350) which tries to delay, but is to small or at the wrong position.

According to the datasheet of the ADS111x:  
* in the section "Output Data Rate and Conversion Time" "Conversions in the ADS111x settle within a single cycle; thus, the conversion time is equal to 1 / DR.", which the above mentioned line of gobot driver code tries to implement
* in the table "Config Register Field Descriptions" the "OS can only be written when in power-down state and **has no effect when a conversion is ongoing.**", which is not implemented yet

A raw implementation of a "wait for bit15 is 1" seems to fix the problem.

I will make further tests and can possibly provide a fix by one of the both approaches or add a With...() function for that.

gen2thomas:
* fix #827 
* work with stable behavior by default (maybe slower)
* introduce a With function to switch to old, maybe the faster approach
* include improvements for #828

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@ ALL_EXAMPLES := $(shell grep -l -r --include "*.go" 'build example' ./)
 # prevent examples with joystick (sdl2) and gocv (opencv) dependencies
 EXAMPLES := $(shell grep -L 'joystick' $$(grep -L 'gocv' $(ALL_EXAMPLES)))
 
-.PHONY: test test_race test_cover robeaux version_check fmt_check fmt_fix examples $(EXAMPLES)
+.PHONY: test test_race test_cover robeaux version_check fmt_check fmt_fix examples examples_check $(EXAMPLES)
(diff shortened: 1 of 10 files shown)


gen2thomas:
convert back to draft, because some new ideas for i2c basic reworks

gen2thomas:
* test coverage improved
* heavily tested with Tinkerboard

PR #831, opened 2022-08-31
BUGFIX: missing HMC5883L byte swap

### LINKED ISSUE #805: I2C driver typo: change HMC8553L to HMC5883L
Thanks @yurigorokhov for #781 but...

It looks like there's a typo regarding driver name.

Should be <code>HMC**5883**L</code> instead of <code>HMC**8553**L</code>.

### LINKED ISSUE #830: HMC5883L driver returns wrong values
According to the datasheet section "Data Output X Registers A and B" the values should be in 2s complement form +0xF800 ... 0x07FF (-2048 ... +2047). In case of an overflow, the value gets "-4096 (0xF000)".

In my first test case, the values of X,Y, Z jump in a high range. All values can exceed 2048 and don't lock to -4096 for a strong magnetic field.

A first investigation shows, that a missing byte-swap after reading the MSB, LSB data with "ReadWordData()" could be the root cause.

I will provide a fix after some further tests, maybe together with #805 .

gen2thomas:
* fix #830 
* add unit tests for reading
* add some constants and With functions for "samples averaged", "data output rate", "applied bias" and "gain"
* fix #805
* use ReadBlockData()


gen2thomas:
@deadprogram now ready to review

gen2thomas:
Set it back to draft, because it will most likely conflict with #833 , but 833 is more important because it is a prerequisite for some more PR's.

gen2thomas:
* before rework, the driver was generally not working right
* the test coverage was heavily improved according to data sheet and before rework was done
* practical test was passed with digispark and Tinkerboard

PR #837, opened 2022-09-15
platforms/ble: support for Windows and remove macOS specific UUID code

deadprogram:
This PR adds BLE support for Windows and also removes macOS specific UUID code no longer needed.


deadprogram:
Replaced by #856

Issue #838, opened 2022-09-18
MPU6050 wrong initialize and reduced temperature resolution, missing convert of accel and gyro to physical units


gen2thomas:
The initialization routine seems to be an adaptation of the [https://github.com/adafruit/Adafruit_CircuitPython_MPU6050](https://github.com/adafruit/Adafruit_CircuitPython_MPU6050/blob/4ec1d2c264816bfb1340b1191b8242abd9b45eb3/adafruit_mpu6050.py#L181) but have some mistakes:
* "RWBits" and "RWBit" working in a different way like "i2c.Write()", which only supports a list of bytes to write, this leads to
 * wrong configuration for PWR1 FS_SEL register (should be 0x01, is 0x00 - "8Ghz internal" should be avoided according to data sheet)
 * accidentally set of of PWR2 register with 0x03 and afterwards 0x00
 * wrong configuration of GYRO register (should be 0x00 is 0x04), but FS_SEL seems to be correct with 0x00
 * wrong configuration of ACCEL register (should be 0x00 is 0x04), but AFS_SEL seems to be correct with 0x00
 * access of undefined registers (0x6D, 0x1D, 0x1E)
* reset of general config register (0x1A) is missing
* the reset call is missing, which is also suggested by the data sheet (DEVICE_RESET and SIGNAL_PATH_RESET) when using the SPI interface

The temperature is stored in °C but only as int16. The resolution is 16bit, so it should be possible to have an resolution of at least "1/340" (according to the data sheets equation). Also the python implementation use a slightly other offset of "+ 36.53", gobots implementation us 12412 / 340 = 36,506.

The acceleration and gyroscopic values are stored as int16 without apply the currently used gain. This leads to the requirement of converting the value at user side. Together with some wrong initialization (see above), this could maybe not possible.

gen2thomas:
part of release v2.0.0

PR #839, opened 2022-09-18
BUGFIX/Improvement: mpu6050 wrong initialize

gen2thomas:
This is for issue #838 
* FIX: initialization sequence
* FIX: temperature resolution to float64
* FIX: Acceleration values now has unit [m²/s]
* FIX: Gyroscope values now has the unit [°/s]
* tests heavily improved
* some new constants introduced
* With-functions for full scale range (gain) and other values introduced
* 2c.Driver used as base, which let us more focus on important tests

### DIFF
--- a/drivers/i2c/mpu6050_driver.go
+++ b/drivers/i2c/mpu6050_driver.go
@@ -3,49 +3,121 @@ package i2c
 import (
 	"bytes"
 	"encoding/binary"
+	"fmt"
+	"log"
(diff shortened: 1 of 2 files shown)


gen2thomas:
* before rework, the driver was not working in most cases
* the test coverage was heavily improved according to data sheet and before rework was done
* practical test was passed with digispark and Tinkerboard

PR #842, opened 2022-09-19
BUGFIX/Improvement: th02 i2c driver - wrong register usage for heater

### LINKED ISSUE #841: TH02 wrong register usage for read heater
TH02 reads the register content of register "0x11" (17) which is the ID register. The heater information is accessible in configuration register  0x03, bit1.

gen2thomas:
This PR fix #841 

In addition:
* add "WithTH02FastMode"
* add set heater independent of measurement
* add mutex to prevent interruption of write-read sequences
* make waitForReady() not so greedy
* tests heavily improved


gen2thomas:
I will going to merge this without further review, because the test coverage was heavily improved according to data sheet and before rework was done.

Issue #832, opened 2022-09-04
i2c/smBus ReadBlockData() not implemented, all "Read*Data()" of digispark and firmata are wrong implemented


gen2thomas:
According to the [Kernel specification](https://docs.kernel.org/i2c/smbus-protocol.html#smbus-block-read) it should be possible to implement this function. The opposite ["WriteBlockData()" is still implemented](https://github.com/hybridgroup/gobot/blob/d8081f912d324f570c6e3fa74aba7c203e53bb27/drivers/i2c/i2c.go#L44).

Following drivers could be simplified:
* [th02_driver.go](https://github.com/hybridgroup/gobot/blob/d8081f912d324f570c6e3fa74aba7c203e53bb27/drivers/i2c/th02_driver.go#L234)
* [mpu6050_driver.go](https://github.com/hybridgroup/gobot/blob/d8081f912d324f570c6e3fa74aba7c203e53bb27/drivers/i2c/mpu6050_driver.go#L100)
* [mpl115a2_driver.go](https://github.com/hybridgroup/gobot/blob/d8081f912d324f570c6e3fa74aba7c203e53bb27/drivers/i2c/mpl115a2_driver.go#L120) and for [getData()](https://github.com/hybridgroup/gobot/blob/d8081f912d324f570c6e3fa74aba7c203e53bb27/drivers/i2c/mpl115a2_driver.go#L156)
* [l3gd20h_driver.go](https://github.com/hybridgroup/gobot/blob/d8081f912d324f570c6e3fa74aba7c203e53bb27/drivers/i2c/l3gd20h_driver.go#L115)
* HMC5883L in ReadRawData()
* [ccs811_driver.go](https://github.com/hybridgroup/gobot/blob/d8081f912d324f570c6e3fa74aba7c203e53bb27/drivers/i2c/ccs811_driver.go#L339), an implementation of read block data, also used in all 3 bmpx80 and bme280 drivers
* [adxl345_driver.go](https://github.com/hybridgroup/gobot/blob/d8081f912d324f570c6e3fa74aba7c203e53bb27/drivers/i2c/adxl345_driver.go#L222)

Some notes for support of block-read: 
* "i2cdetect -F 0" show the state
* Raspi Kernel version 4.9.35 [related bug](https://github.com/raspberrypi/firmware/issues/828)
* Tinkerboard "Tinker_Board-Debian-Buster-v3.0.11-20211026" no support
* digispark (using littlewire) support it since #833 
[...]

PR #843, opened 2022-09-20
Feature: MPL115A2 use ReadBlockData()

gen2thomas:
this is part of issue #832 

in addition to switch to ReadBlockData():
* switch also to WriteByteData()
* unexpose some constants and variables
* tests for Start() and ReadData() improved
* add example for Tinkerboard


gen2thomas:

* practical test was passed with Tinkerboard

PR #845, opened 2022-09-21
BUGFIX: L3GD20H full scale range usage

### LINKED ISSUE #844: L3GD20H scale is not used without restart - wrong scale used
Change of scale must be transferred to the device. This is currently done in the initialization routine (called by Start()). If the scale is set with the provided SetScale() it will modify the used scale for next measurement call by "XYZ()". This cause wrong values, because the real used scale was not changed.

In addition:
* The full scale range is 245 instead of used 250 in the constant name. This seems to be just a cosmetically error, because the real used scale is 0.00875, which corresponds to the data sheet table3 page 10 (8.75 mdps/digit).
* The default value of scale of the used switch/case is "0", which leads to an return value of "0" - an error should be returned in this case or in case of "0x20" this leads to dps2000 (70 mdps/digit) according to data sheets "0x1x" definition.

gen2thomas:
This PR fix #844 

In addition:
* FullScaleRange() read the value from device
* use ReadBlockData() & WriteByteData()
* tests heavily improved
* 2c.Driver used as base, which let us more focus on important tests

### DIFF
--- a/drivers/i2c/l3gd20h_driver.go
+++ b/drivers/i2c/l3gd20h_driver.go
@@ -3,43 +3,53 @@ package i2c
 import (
 	"bytes"
 	"encoding/binary"
-
-	"gobot.io/x/gobot"
+	"log"
 )
 
-const l3gd20hAddress = 0x6B
-
-// Control Register 1
-const l3gd20hRegisterCtl1 = 0x20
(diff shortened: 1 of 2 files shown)


gen2thomas:
I will going to merge this without further review, because the test coverage was heavily improved according to data sheet and before rework was done.

PR #846, opened 2022-09-21
Feature: CCS811 use ReadBlockData()

gen2thomas:
this is part of issue #832 

In addition:
* tests improved


gen2thomas:
* no feature implementation was changed

PR #847, opened 2022-09-21
Feature: BMP180 BMP280 BMP388 BME280 use ReadBlockData()

gen2thomas:
this is part of issue #832 

In addition:
* WriteByteData() is used
* "With" functions added
* tests improved
* add example for tinkerboard with BME280 and "With"-function usage for indoor navigation


gen2thomas:

* practical test was passed with Tinkerboard

PR #856, opened 2022-09-25
windows: BLE support

deadprogram:
This PR adds BLE support for Windows!


deadprogram:
Merging so we can continue from `dev` branch.

Issue #793, opened 2021-05-25
Dji Tello Halt does not terminate all the related goroutines and may wait forever when it is called multiple times


st-user:
(parent commit [27c76b4585807e829965ec6a4551cadc28f430fe](https://github.com/hybridgroup/gobot/commit/27c76b4585807e829965ec6a4551cadc28f430fe))

I tried Dji Tello Driver and found the following two problem related to `Halt` method.

If these are issues that need to be fixed, I’d like to make a pull request.

### 1. Halt stops only one of the three infinite loops.
 
The Driver runs the **three** infinite loops(*). However `Halt` stops only one of the three because:

- `case <- d.doneCh: break…` is defined in **two** of the three.
- `Halt` calls `d.doneCh <- struct{}{}` only once.

(*)
  - https://github.com/hybridgroup/gobot/blob/release/platforms/dji/tello/driver.go#L289
  - https://github.com/hybridgroup/gobot/blob/release/platforms/dji/tello/driver.go#L308
  - https://github.com/hybridgroup/gobot/blob/release/platforms/dji/tello/driver.go#L949

So, only one of the three infinite loops can read `d.doneCh` and stop.

 ### 2. Halt may wait forever when it is called multiple times.

Simply because `Halt` is blocked by `d.doneCh <- struct{}{}`, if we call `Halt` after the driver has been halted, It blocked forever.

### Test case

The test cases below simulate the problems.
These two test methods don't stop.
(https://github.com/st-user/gobot/commit/1ffc13879c54a4625591add72b7ccf74661e3771)

```
// Only output one line ('Done routine x.')
func TestHaltShouldTerminateAllTheRelatedGoroutines(t *testing.T) {
	d := NewDriver("8888")
	d.cmdConn = &WriteCloserDoNothing{}

	var wg sync.WaitGroup
	wg.Add(3)
	go func() {
		<-d.doneCh
		wg.Done()
		fmt.Println("Done routine 1.")
	}()
	go func() {
		<-d.doneCh
		wg.Done()
		fmt.Println("Done routine 2.")
	}()
	go func() {
		<-d.doneCh
		wg.Done()
		fmt.Println("Done routine 3.")
	}()

	d.Halt()
	wg.Wait()
}

func TestHaltNotWaitForeverWhenCalledMultipleTimes(t *testing.T) {
	d := NewDriver("8888")
	d.cmdConn = &WriteCloserDoNothing{}

	d.Halt()
	d.Halt()
	d.Halt()
[...]

PR #858, opened 2022-09-25
dji/tello: fix the issue about Halt method

st-user:
This PR is intended to fix the issue: [Dji Tello Halt does not terminate all the related goroutines and may wait forever when it is called multiple times #793](https://github.com/hybridgroup/gobot/issues/793).

This PR contains consecutive four commits.


deadprogram:
Thank you very much for this PR @st-user now merging.

PR #863, opened 2022-10-02
Introduce a generic i2c driver

gen2thomas:
This driver can be used as is for simple devices, see provided example.
In addition it is possible to use it as a base of new i2c drivers.

In addition:
* README files fixed
* i2c.go renamed to i2_connection.go
* tests for i2c_*.go files added/improved
* "TestRobotAutomationFunctions" stabilized by current state of #856


gen2thomas:
Because this is a basic step for some following driver improvements and the new driver is tested with the provided digispark example, I going to merge this PR without further review.

PR #866, opened 2022-10-17
modify NewPWMPin function for creation of diffrent PWM chip number

rich1111:
Hi, i modify NewPWMPin function for the need of creation of different PWM chip number.


gen2thomas:
Hi @rich1111 , thanks for contributing to gobot. Please can you change the [target branch to dev](https://github.com/hybridgroup/gobot/blob/release/CONTRIBUTING.md#target-branch). I will have a deeper look to your changes this week.

Do you have a real need/application/platform for this change? Do you have seen, that some platforms override the path (using a "findPin()" function)?

Issue #849, opened 2022-09-24
CI/CD environment outdated


gen2thomas:
* deprecated  (circleci), see https://discuss.circleci.com/t/legacy-convenience-image-deprecation/41034
* golang [1.15 is used](https://github.com/hybridgroup/gobot/blob/f141a2bc7a562f500ced5ab679bede9816fa01f1/.circleci/config.yml#L6), which cause CI build errors for new functions of golang 1.16 (e.g. os.ReadDir), see https://app.circleci.com/pipelines/github/hybridgroup/gobot/40/workflows/ef2f4957-0873-4dd8-8784-dfe5c9348e12/jobs/40

deadprogram:
Fixed!

gen2thomas:
Thanks for fast merging.

Mhm, unfortunately the change from /go to $HOME/go in new cimg seems to make some problems, see
https://app.circleci.com/pipelines/github/hybridgroup/gobot/41/workflows/dfaa8bbf-48ed-4b6e-b91b-bbceedd52bba/jobs/41?invite=true#step-101-73

Is there a reason why we see this problems only on "dev"? - Yes, because we run the circleci not on feature branches, only the "appveyor" build. 

Is this done by intention?

The "appveyor" build and test is done on a windows environment. Some tests can most likely not run on this platform:
* the sysfs-tests (generic access to linux gpio)
* all adaptor tests which depends on sysfs (means all, except firmata)

gen2thomas:
After merging the last fix, tests will be start. But following tests now fail:
* "utils_test.go:38: Every was not called"
* "utils_test.go:44: Every should have stopped"

Because this tests running before without any problems, it seems to be flaky.

Also this test seems to be flaky:
*  "client_test.go:466: SysexResponse was not published"

deadprogram:
Yes, they appear to be very sensitive to race conditions/timing.

gen2thomas:
still unstable:
* [core](https://app.circleci.com/pipelines/github/hybridgroup/gobot/68/workflows/9f80bfb8-d832-4e19-8c26-262ce31b116c/jobs/75?invite=true#step-103-70)

* [platforms](https://app.circleci.com/pipelines/github/gen2thomas/gobot/8/workflows/30de7e0f-203f-4529-a5ad-87fd10c02ee1/jobs/16?invite=true#step-103-400)

gen2thomas:
part of release v2.0.0

PR #867, opened 2022-10-18
TESTFIX: stabilize firmata tests

gen2thomas:
this tests often fail, see also #849 :
* TestProcessSysexData, client_test.go:466: SysexResponse was not published OR panic: test timed out after 10m0s
* TestConnect

root causes are:
* read/write access to one resource from different tests (which run in parallel): "testReadData"
* initialization with more than needed data
* used mutexs can lock each other

### DIFF
--- a/platforms/firmata/client/client_test.go
+++ b/platforms/firmata/client/client_test.go
@@ -2,6 +2,7 @@ package client
 
 import (
 	"bytes"
+	"log"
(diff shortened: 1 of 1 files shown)


gen2thomas:
going to merge without further review, to enable new test behavior asap in daily use (no production code changed)

PR #871, opened 2022-10-21
Joystick add Xbox-One controller

gen2thomas:
based on the work by @willmendil (#757) and review by @krobertson


gen2thomas:
basic work already reviewed

PR #879, opened 2022-10-26
BUGFIX: tests with sysfs mocks

### LINKED ISSUE #878: Tests, which using sysfs mocks are/can conflicting each other
While working on the driver for PCA9533 I have recognized, that there is a bug in i2c.WriteBlockData() for sysfs implementation. To show this in a unit test I tried to write one and recognized, that the tests in [i2c_device_test.go](https://github.com/hybridgroup/gobot/blob/dev/sysfs/i2c_device_test.go) are not independent (means can not be run in parallel or when running sequential are order dependant), because the usage of the same resource "sysfs.fs" and "sysfs.syscall", which is set to a mocked system. There are 2 problems with the current approach:
* after setting the variable it is not reset after the test is finished (also affects tests when not running in parallel)
* one test can set this variable(s) and another can use it (when running in parallel)

The first step will be to cleanup after each test by using a defer function. This ensures independently running for sequential call and do not affect the implementation itself. I'm currently working on it.

To ensure the tests can be running in parallel (without becomes flaky), some changes in implementation are needed.

gen2thomas:
this fixes #878 part 1 and:
* add README on sysfs for developers
* improve test heavily
* try prevent uintptr as long as possible
* fix ReadBlockData(), WriteBlockData()
* wrong comments in examples for PCF8583
* provide connection functions in generic i2c_driver.go

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,16 +13,18 @@ EXAMPLES := $(EXAMPLES_NO_GOCV)
 including_except := $(shell go list ./... | grep -v platforms/opencv)
 
 # Run tests on nearly all directories without test cache
+# TODO: set back to parallel running (remove "-p 1"), when issue #878 is fixed
 test:
-	go test -count=1 -v $(including_except)
+	go test -p 1 -count=1 -v $(including_except)
 
 # Run tests with race detection
 test_race:
 	go test -race $(including_except)
 
 # Test, generate and show coverage in browser
+# TODO: set back to parallel running (remove "-p 1"), when issue #878 is fixed
 test_cover:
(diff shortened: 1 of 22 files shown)


gen2thomas:
Change for unsafe.Pointer and WriteBlockData(), ReadBlockData() tested with this devices and tinkerboard:
* PCA9501 (only writes)
* ADXL345
* BME280
* BMP280 (with attached BME280)
* MPL115A2
* ADS1115
* PCF8583 (also temporary changed to use formerly not working WriteBlockData())

gen2thomas:
I will going to merge without further review because all hardware and automated tests passed.

PR #821, opened 2022-05-29
Added support for Dualsense joystick (PlayStation 5)

jig:
Added support for PlayStation 5 Dualsense joystick.

Code derived from Dualshock 4:
- buttons remapped to appropriate IDs
- L2 and R2 are now "axis" instead of "buttons"
- New "create" button (instead of "share" button)
- examples/joystick_ps5.go file as example code

PR #880, opened 2022-10-29
Added support for Dualsense joystick (PlayStation 5)

jig:
Added support for PlayStation 5 Dualsense joystick.

Code derived from Dualshock 4 implementation:

* buttons remapped to appropriate IDs
* L2 and R2 are now "axis" instead of "buttons"
* New "create" button (instead of "share" button)
* look at examples/joystick_ps5.go file as a usage example

This PR supersedes https://github.com/hybridgroup/gobot/pull/821.

### DIFF
--- a/examples/joystick_ps5.go
+++ b/examples/joystick_ps5.go
@@ -1,0 +1,140 @@
+//go:build example
+// +build example
+
+//
+// Do not build by default.
+
+package main
+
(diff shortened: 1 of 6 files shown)


gen2thomas:
Hi @jig , thanks for the continued work. I'm going to merge the PR now.

PR #881, opened 2022-11-01
BUGFIX: tests with sysfs mocks, part 2

### LINKED ISSUE #878: Tests, which using sysfs mocks are/can conflicting each other

gen2thomas:
this fixes #878 part 2, and additionally:

* unexpose some sysfs items
* introduce sysfs.ReadFile() and use for read revision of raspi
* add a README.md for sysfs
* apply [YAGNI principle](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) on some adaptors (no change of interface)
* improve tests a lot
* usage of writeFile, readFile "func" in some adaptors substituted by sysfs mock
* refute-nil-assertions in tests for errors - replaced with string test when useful
* tests for GetDefaultBus() etc. in adaptors
* fix BlockWriteData/BlockReadData & improve new README.md
* add tinkerboard examples for ADXL345, HMC5883L, MPU6050
* pcf8583_driver now working with WriteBlockData() - changed
* final test with Tinkerboard and some devices
* final test with raspi and PCA9501
* activate parallel tests in Makefile
* update README.md of Tinkerboard because discontinued "Tinker OS"


gen2thomas:
Run HW tests because usage of ReadBlockData or WriteBlockData on some of those devices:
* adxl345_driver - OK
* bme280_driver, bmp180_driver, bmp280_driver, bmp388_driver - OK
* hmc5883l_driver - OK
* l3gd20h_driver - n.t.
* mpl115a2_driver - OK
* mpu6050_driver - OK
* th02_driver - n.t.
* pcf8583_driver (now working with WriteBlockData) - OK
* ccs811_driver - n.t.
* PCA9501 EEPROM is finally working with raspi and on tinkerboard with installed armbian

gen2thomas:
* "golint" all changed files
* own PR review

gen2thomas:
going to merge the PR without further review because heavily tested and unit tests improved

PR #885, opened 2022-11-20
rename package "sysfs" to "system"

gen2thomas:
This PR prepares the implementation of character devices for GPIO and clean up some uncommon names and structures.
* rename package "sysfs" to "system", because it provides access to some system features (not only sysfs anymore)
* rename local variables "sysfs" to "sys"
* move interfaces for DigitalPin and PWMPin away from implementation
* use interfaces for local pin-caches instead of pointer to implementation (precondition for use different implementations)
* remove still unused custom label option from sysfs-DigitalPin
* unexpose Path from pwmpin


gen2thomas:
merge without further review, because just refactoring without untested code changes

Issue #775, opened 2020-12-27
Add support for GPIO Character devices on Linux (specifically raspberry pi)


joeblubaugh:
[The sysfs interface to gpio is deprecated](https://embeddedbits.org/linux-kernel-gpio-user-space-interface/) and will be removed from the kernel sometime in 2020.

The kernel now has character device support for GPIO. Here's the output from some gpio commands on my Raspberry Pi 400:
```
pi@raspberrypi:/dev $ gpioinfo 
gpiochip0 - 54 lines:
        line   0:     "ID_SDA"       unused   input  active-high 
        line   1:     "ID_SCL"       unused   input  active-high 
        line   2:       "SDA1"       unused   input  active-high 
        line   3:       "SCL1"       unused   input  active-high 
        line   4:  "GPIO_GCLK"       unused   input  active-high 
        line   5:      "GPIO5"       unused   input  active-high 
        line   6:      "GPIO6"       unused   input  active-high 
        line   7:  "SPI_CE1_N"       unused   input  active-high 
        line   8:  "SPI_CE0_N"       unused   input  active-high 
        line   9:   "SPI_MISO"       unused   input  active-high 
        line  10:   "SPI_MOSI"       unused   input  active-high 
        line  11:   "SPI_SCLK"       unused   input  active-high 
        line  12:     "GPIO12"       unused   input  active-high 
        line  13:     "GPIO13"       unused   input  active-high 
        line  14:       "TXD1"       unused   input  active-high 
        line  15:       "RXD1"       unused   input  active-high 
        line  16:     "GPIO16"       unused   input  active-high 
        line  17:     "GPIO17"       unused   input  active-high 
        line  18:     "GPIO18"       unused   input  active-high 
        line  19:     "GPIO19"       unused   input  active-high 
        line  20:     "GPIO20"       unused   input  active-high 
        line  21:     "GPIO21"       unused   input  active-high 
        line  22:     "GPIO22"       unused   input  active-high 
        line  23:     "GPIO23"       unused   input  active-high 
        line  24:     "GPIO24"       unused   input  active-high 
        line  25:     "GPIO25"       unused   input  active-high 
        line  26:     "GPIO26"       unused   input  active-high 
        line  27:     "GPIO27"       unused   input  active-high 
        line  28: "RGMII_MDIO"       unused   input  active-high 
        line  29:  "RGMIO_MDC"       unused   input  active-high 
        line  30:       "CTS0"       unused   input  active-high 
[...]

Issue #565, opened 2018-06-10
Stopping and Starting Robot with Raspberry Pi Adapter and LED Driver - LED does not toggle on restart


acarlonvs:
**Summary**
Starting, Stopping and Starting robot with Raspi Adapter and LED driver does not work on the second Start(). The LED does not toggle and Toggle returns "pin has not been exported".

**Cause**
Given the sample code from [https://gobot.io/documentation/platforms/raspi/](url): 

```
func main() {
        r := raspi.NewAdaptor()
        led := gpio.NewLedDriver(r, "7")
        work := func() {
                gobot.Every(1*time.Second, func() {
                        led.Toggle()
                })
        }
        robot := gobot.NewRobot("blinkBot",
                []gobot.Connection{r},
                []gobot.Device{led},
                work,
        )
        robot.Start()
}
```
If the robot is stopped and then restarted, the LED does not blink. The reason is in raspi_adapter.go. When the robot is stopped raspi_adapter.go `Finalize()` will `Unexport` the pins, but the pins are retained in the map. So, when the robot is again started and `DigitalWrite() -> DigitalPin() -> getExportedDigitalPin()` is called, we get to the following code: 

```
        if r.digitalPins[translatedPin] == nil {
		r.digitalPins[translatedPin] = sysfs.NewDigitalPin(translatedPin)
		if err = r.digitalPins[translatedPin].Export(); err != nil {
			return
		}
	}
```
Note, the pin is still in the map, but is now unexported. This results in the "pin has not been exported" error.

**Fix** 

First remove the map initialization from `NewAdaptor()`
```
 func NewAdaptor() *Adaptor {
        r := &Adaptor{
-               mutex:       &sync.Mutex{},
-               name:        gobot.DefaultName("RaspberryPi"),
-               digitalPins: make(map[int]*sysfs.DigitalPin),
-               pwmPins:     make(map[int]*PWMPin),
+               mutex: &sync.Mutex{},
+               name:  gobot.DefaultName("RaspberryPi"),
 }
```
Second, add to `Connect()`: 
```
 func (r *Adaptor) Connect() (err error) {
+       r.digitalPins = make(map[int]*sysfs.DigitalPin)
+       r.pwmPins = make(map[int]*PWMPin)
        return
 }
```
This will ensure that a new map is used each time the Robot is started. Not sure if there are any unintended effects of doing this, but at least it highlights what the problem is. 
I can try to do a pull request on dev with this change if that is helpful.

[...]

Issue #439, opened 2017-07-18
raspberry pi built-in pull up/down resistor support?


boundedinfinity:
Is there a way to configure the raspberry pi internal pull up/down resistors?  I'm looking for help (or just a pointer to documentation) do some something similar to the configuration of the GPIO pin's pull up/down resistor in the python [RPi.GPIO](https://sourceforge.net/p/raspberry-gpio-python/wiki/BasicUsage/) module.

In python the configuration is achieved by something like this:

```python
GPIO.setup(PIN, GPIO.IN, pull_up_down = GPIO.PUD_DOWN)
```

Does gobot have something similar?

spatil:
Is there any workaround for this?

gokuu:
I have the same issue :(

stuartleeks:
@deadprogram - do you have any sense of what would be needed to add this? (From what I can tell it isn't currently supported?)

gen2thomas:
part of release v2.0.0

Issue #416, opened 2017-05-05
GPIO Interrupt Support


pauleyj:
Any thoughts or plans on supporting GPIO interrupts?

larsth:
Epoll is your friend:

https://godoc.org/syscall#EpollCreate
https://godoc.org/syscall#EpollCreate1
https://godoc.org/syscall#EpollCtl
https://godoc.org/syscall#EpollWait

epoll(7) man page: http://man7.org/linux/man-pages/man7/epoll.7.html
epoll(2) man page: https://linux.die.net/man/4/epoll

Use EPOLLET mode / flag when using EpollCreate, EpollCreate1

mikegleasonjr:
So should we implement this in userland and not user specific drivers?

deadprogram:
Hi everyone. Somehow I previously missed @larsth comment when it was originally made. Yes, epoll would be a great solution. Seems like by adding some wrapper code to the Gobot sysfs package, getting to work on any Linux such as Raspi, Beaglebone Black, etc. would be straightforward.

@larsth if there is a kernel module that blocks the interrupt, that probably means a kernel driver that is already using it, and so it would be best for Gobot to leave it alone, right? What is the correct way to handle this?

@mikegleasonjr yes, that seems like the most likely pattern for a Gobot implementation. I personally avoid adding kernel modules unless absolutely needed for some special purpose.

larsth:
> @larsth if there is a kernel module that blocks the interrupt, that probably means a kernel driver that is already using it, and so it would be best for Gobot to leave it alone, right? 

Yes

> What is the correct way to handle this?

The only thing that can be done is to remove the module from the kernel:
`sudo rmmod >name of the offending kernel module goes here<`

What happens after that command shows whether or not removing the module from the kernel was a good idea, however, I think it would be safe to remove the module, because the user of Gobot is probably doing something else with the GPIO pins.

Controlling a piece of hardware with only one GPIO pin at a time is highly undesirable because context switches creates long delays. 
If only there was a GPIO kernel module with poll syscall support it would be great, but I don't have any time anytime soon to create such a module. The 2 times earlier that I had made a character device module was very fun. It was 9 years ago.

BTW,
It is fun to watch that after fx a shell command like this:
`echo 1 > /dev/my_character_device_file` 
... there continues to be action, because inertia prevents anything to happen instantly.

gen2thomas:
part of release v2.0.0

PR #887, opened 2022-11-27
gpio: add support for new character device Kernel ABI

gen2thomas:
this implementation fixes the following issues:
* #775 
* #565 (partially for digital pins)

it opens the implementation for following issues:
* #439 
* #416 

Additionally:
* new feature is now used in "raspi" and "tinkerboard" adaptors
* if character devices are not supported by the OS, the fallback is sysfs (automatically)
* introduce a DigitalPinAdaptor to composite the platforms with it, to reduce (duplicated) lines of code
* introduce options for digital pins, also usable for existing sysfs implementation

Interface changes:
* setter function "Direction(string) error" removed from "gobot.DigitalPinner", now done by "WithDirection..." option
* getter function "DigitalPin(string, string)" in "gobot.DigitalPinnerProvider" do not change the direction anymore, it provides the pin as is and if not exist it will be created as input, changes can be applied afterwards with "ApplyOptions()", to prevent accidentally direction changes, which could cause board damages


gen2thomas:
I'm going to merge this without further review, because:
* heavily tested manually for raspi and Tinkerboard
* unit test coverage 80-95% for the new code (except the reconfigure part for gpiod, which was tested manually)
* mandatory precondition for further bugfixes and improvements

PR #890, opened 2022-12-03
Bugfix/Improvement: Use PWMPinsAdaptor for platforms

gen2thomas:
Fix the issue #565 for PWM pins.

Following platforms are adapted:
- [x] beaglebone (mux pin)
- [x] chip
- [x] edison
- [x] joule (with own "pwmPinInitializer")
- [x] jetson (own PWM impl. but cache fixed)
- [x] raspi, dropped, because use pi-blaster
- [x] tinkerboard
   - [x] manual test
- [x] up2 (LED's)

additionally fixes:
* access of "enable" if period is 0 leads to errors (affected: Tinkerboard, chip, up2)
* Tinkerboard: fix usage in armbian if only one pwm is used (disabling is possible in armbian via device tree)
* ServoWrite() now uses the real period (affected: beaglebone, chip, jetson)
* jetson has dropped all errors on ServoWrite() 

Interface changes:
* Enable() --> SetEnabled() and introduce Enabled() getter, because better match the golang style
* SetPolarity(string) --> SetPolarity(bool) and remove InvertPolarity(bool), because double inversion should be avoided
* Polarity() string --> Polarity() bool, because symmetric implementation of getter


gen2thomas:
I'm going to merge without further review, because:
* test coverage improved
* additional bugs fixed (now better than before)
* manual test with Tinkerboard

PR #894, opened 2022-12-10
Improve: use base driver for all I2C devices

gen2thomas:
Use the already existent base i2c_driver also for all remaining drivers, if possible. This reduces the lines of code and simplifies all tests. New tests can be now focused on the real driver implementations.

Interface changes to improve  visual differentiation on adaptor level:
* DefaultBus() --> DefaultI2cBus()
* GetConnection() --> GetI2cConnection()

Interface changes to be more conform with golang style:
* WithBus() --> SetBus()
* WithAddress() --> SetAddress()

Also moved the interfaces to client, to be more conform with golang style.


gen2thomas:
Merge without further review, because heavily tested and lines of code reduced.

PR #898, opened 2023-02-04
BUGFIX: I2C connection-bus caching and multiple device usage

gen2thomas:
rollback previous change on dev branch for I2C connection caching (bus needs to be cached and closed by adaptor, not the connection)

fixes that sequence SetAddress(A) --> CommandFor(A) can now not be interrupted anymore, e.g. by
SetAddress(A) -.....................................................................-> CommandFor(A)
..........................\ ->SetAddress(B) --> CommandFor(B)  -/

which leads to wrong actions or bus interruption

furthermore:
* set address only when changed on bus
* fix some small golint errors

### DIFF
--- a/adaptor.go
+++ b/adaptor.go
@@ -93,10 +93,67 @@ type PWMPinnerProvider interface {
 	PWMPin(id string) (PWMPinner, error)
 }
 
-// I2cSystemDevicer is the interface to a i2c bus at system level.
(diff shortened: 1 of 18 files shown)


gen2thomas:
I'm going to merge without external review, because:
* reverts formerly change to its original state
* fixes instability
* manually tested with Tinkerboard and 2 I2C devices on bus

PR #902, opened 2023-02-10
Add support for Radxa's Rock Pi 4(C+) Single Board Computers

brainbaking:
Hi,

I've been fiddling with a Raspberry Pi clone for a while and managed to get CircuitPython running on it. I've contributed the correct pinout here this week: https://github.com/adafruit/Adafruit_Blinka/pull/651
But since I wanted to do something in Go with I2C, I stumbled upon Gobot, which doesn't have the correct mapping either. Here's my attempt to add compatibility in Gobot. 

Please note that, although I've read the CONTRIBUTING guide, I am unable to run unit tests thanks to a nasty dependency: https://github.com/warthog618/gpiod which literally states: "For other platforms, where you intend to cross-compile for Linux, don't attempt to compile the package when it is installed". On my Mac, I get a lot of errors like "../../../../go/pkg/mod/github.com/warthog618/gpiod@v0.8.1/gpiod.go:326:35: undefined: uapi.LineInfo". On the RockPi itself, the Go version isn't recent enough. 
But since the `rockpi_adaptor.go` is based on the Pi's, and it has reduced functionality, and the most important aspect is the mapping, I figured this is as good as it gets. 

Note that Radxa's boards aren't great when it comes to support/compatibility with existing software. The character device gpiod driver doesn't seem to work, and PWM only works through the `mraa` lib, which could be implemented using `cgo`, but I didn't yet. I think that's for another contributor!

Thanks for any feedback! Cheers


gen2thomas:
Hi @wgroeneveld , thanks for sharing your nice work. I will have a deeper look in the next days. What I can say immediately:
* please can you add some of your description of the PR to the README.md, e.g. in a section I2C, PWM, GPIO etc.
* if the gpiod chracter device is known not to work, just remove the "system.WithDigitalPinGpiodAccess()" in the constructor (line 52)

brainbaking:
Great, thanks, I did that, and I managed to add tests by locally removing the problematic dependency!

platforms/rockpi/README.md
```diff
@@ -0,0 +1,86 @@
...
+        )
+
+        robot.Start()
+}
+```
+
+If you want to use I2C, RockPi4 offers three I2C buses: I2C2 (pins 27, 28), I2C6 (pins 29, 31) and I2C7 (pins 3, 5) of which I2C7 is the default.
+Changing this is a matter of passing the right bus number:
+
+```go
```
gen2thomas, line 62: practically this idiom is used (the default address of ads1115 is 0x48):
adc2 := i2c.NewADS1115Driver(a, i2c.WithAddress(0x49))

platforms/rockpi/rockpi_adaptor.go
```diff
@@ -0,0 +1,165 @@
...
+
+// NewAdaptor creates a RockPi Adaptor
+// Do not forget to enable the required overlays in /boot/hw_initfc.conf!
+// See https://wiki.radxa.com/Rockpi4/dev/libmraa
+//
+// Optional parameters:
+//
+//	adaptors.WithGpiodAccess():	use character device gpiod driver instead of the default sysfs (does NOT work on RockPi4C+!)
+//	adaptors.WithSpiGpioAccess(sclk, nss, mosi, miso):	use GPIO's instead of /dev/spidev#.#
+//	adaptors.WithGpiosActiveLow(pin's): invert the pin behavior
```
gen2thomas, line 46: without character device driver this will most likely not work (except "active low")

platforms/rockpi/rockpi_adaptor.go
```diff
@@ -0,0 +1,165 @@
...
+	return func(pin string) (chip string, line int, err error) {
+		if val, ok := pins[pin][c.readRevision()]; ok {
+			line = val
+		} else if val, ok := pins[pin]["*"]; ok {
+			line = val
+		} else {
+			err = errors.New("Not a valid pin")
+			return
+		}
+		// The chipname is only used in char device GPIO access, which is not used by default.
```
gen2thomas: Please have a look at chip_adaptor.go as one example for classic translator for sysfs. Maybe this example will not fulfill all requirements for this driver. What I would like to say: If the character device support is currently not working (and is not tested with the adaptor), we should not implement anything for that.

platforms/rockpi/rockpi_adaptor_test.go
```diff
@@ -0,0 +1,136 @@
...
+)
+
+func initTestAdaptorWithMockedFilesystem(mockPaths []string) (*Adaptor, *system.MockFilesystem) {
+	a := NewAdaptor()
+	fs := a.sys.UseMockFilesystem(mockPaths)
+	a.Connect()
+	return a, fs
+}
+
+func TestI2cDefaultBus(t *testing.T) {
```
gen2thomas:
```suggestion
func TestDefaultI2cBus(t *testing.T) {
```

platforms/rockpi/rockpi_adaptor_test.go
```diff
@@ -0,0 +1,136 @@
...
+	a, _ := initTestAdaptorWithMockedFilesystem([]string{})
+	gobottest.Assert(t, a.DefaultI2cBus(), 7)
+}
+
+func TestGetPinTranslation(t *testing.T) {
```
gen2thomas:
```suggestion
func Test_getPinTranslatorFunction(t *testing.T) {
```

platforms/rockpi/rockpi_adaptor_test.go
```diff
@@ -0,0 +1,136 @@
...
+			// act
+			_, line, err := fn(tc.pin)
+			// assert
+			gobottest.Assert(t, err, tc.expectedErr)
+			gobottest.Assert(t, line, tc.expectedLine)
+		})
+	}
+}
+
+func TestValidateSpiBusNumber(t *testing.T) {
```
gen2thomas:
```suggestion
func Test_validateSpiBusNumber(t *testing.T) {
```

platforms/rockpi/rockpi_adaptor_test.go
```diff
@@ -0,0 +1,136 @@
...
+			a := NewAdaptor()
+			// act
+			err := a.validateSpiBusNumber(tc.busNr)
+			// assert
+			gobottest.Assert(t, err, tc.expectedErr)
+		})
+	}
+}
+
+func TestValidateI2cBusNumber(t *testing.T) {
```
gen2thomas:
```suggestion
func Test_validateI2cBusNumber(t *testing.T) {
```

platforms/rockpi/rockpi_adaptor_test.go
gen2thomas: reason: we try to use this wording:
* `Test<MethodName>` for exposed functions
* `Test_<methodName>` for unexposed functions

platforms/rockpi/README.md
```diff
@@ -0,0 +1,86 @@
...
+- Rock Pi 4
+- Rock Pi 4C+
+
+With the possibility to expand its compatibility into past and future models.
+
+Check out the output of `cat /proc/device-tree/model` to see which model you have if you're not sure. The 4C+ model has a Rockchip 3399_T SoC, while the regular 4 has the 3399. Both are similar, but have slightly different GPIO pin configurations.
+
+## How to Install
+
+Make sure you've installed an official Linux image from Radxa with working drivers. Some versions or Armbian ISOs do not detect the newer SoC chips! See the [ROCK 4 Installation Wiki](https://wiki.radxa.com/Rock4/install) for your SBC setup.
```
gen2thomas:
```suggestion
Make sure you've installed an official Linux image from Radxa with working drivers. See the [ROCK 4 Installation Wiki](https://wiki.radxa.com/Rock4/install) for your SBC setup.
> Some versions of Armbian ISOs do not detect the newer SoC chips.
```

gen2thomas: There are some images available at https://mirror-eu-de1.armbian.airframes.io/dl/rockpi-4cplus/archive/. Do you have tested this? The main advantage of this images: the newest kernel with latest fixes and character device support for gpio and the newest Debian/Ubuntu. IMO we should use this instead of the now unmaintained images of the SoC provider due to security issues and lack of features. But I know the problem with RockChip-based SoC-support for armbian - I want to use Tinkerboard2 (RK3399) with armbian too and it has not worked in Dec 2022.

gen2thomas:
Hi @wgroeneveld I have finished the PR review for now. For your remark 
> On my Mac, I get a lot of errors like "../../../../go/pkg/mod/github.com/warthog618/gpiod@v0.8.1/gpiod.go:326:35: undefined: uapi.LineInfo". On the RockPi itself, the Go version isn't recent enough.

We use v0.8.0 in gobot and I had not tested the v0.8.1 yet on my Linux developer machine. If your problem persists, please file an issue at [gpiod](https://github.com/warthog618/gpiod/issues), because it is most likely not caused by gobot.

platforms/rockpi/rockpi_adaptor.go
gen2thomas: (the comment is related to lines after line 46)

platforms/rockpi/README.md
gen2thomas: (the comment is related to lines after line 60)

brainbaking: In your example, I can't see where the busnr is set? I wasn't talking about the I2C device address but the I2C buses of the Rockchip 3399_T.

platforms/rockpi/rockpi_adaptor.go
brainbaking: Ok, I'll remove those!

platforms/rockpi/rockpi_adaptor.go
brainbaking: But NewDigitalPinsAdaptor requires a pintranslatorfunction that returns it like this. I checked the raspi_adaptor.go implementation and it also has its shortcomings (see the TODO there at line 197), so I'm not sure what you want me to do here?

platforms/rockpi/README.md
brainbaking: I have tested them all and none work except the official Radxa debian image.

brainbaking:
Thank you for your review @gen2thomas , I've changed the files as requested.
I know the gpiod issue isn't caused by gobot, see my previous comments.

platforms/rockpi/README.md
gen2thomas: Oh, sorry, than it is: adc2 := i2c.NewADS1115Driver(a, i2c.WithBus(6)) for example

platforms/rockpi/rockpi_adaptor.go
gen2thomas: The simplest solution is to return an empty string "" instead of a not existent chip. Afterwards the comment can be removed, because this is already mentioned in the README.

platforms/rockpi/README.md
brainbaking: Right, I'll change it!

platforms/rockpi/rockpi_adaptor.go
brainbaking: OK thanks I'll change it.

brainbaking:
Okay I solved the two open review issues, I think everything looks good now!

gen2thomas:
Yes, thanks for your rework. I will going to merge now.

PR #903, opened 2023-02-17
Add configuration for Nintendo Switch controllers

adregner:
These ID's are what the test program printed out for all the buttons when I had a pair of joycon's connected with Bluetooth to my Mac.  Some of them seem to change when there's only one connected, but they mostly seem like logical changes given the way they are normally held in one- and two-at-once styles.

This is only intended to be 100% accurate for the two-controller style because that's what I'm using them for in my current project.  Thought this could be useful for others that want to repurpose Nintendo Switch controllers too.

While preparing this, I also discovered that some mentions of the Nvidia Shield controller were removed, which I also fixed.


platforms/joystick/joystick_nintendo_joycon.go
```diff
@@ -0,0 +1,115 @@
+package joystick
+
+var joyconPairConfig = joystickConfig{
+	Name: "Nintendo Switch Joycon Controller Pair",
+	GUID: "5555",
```
adregner, line 5: Not sure where to get this, happy to investigate more with some pointers.  In other examples I looked at it seemed random values were being used and I didn't see GUID mentioned in the rest of the codebase anyways.

platforms/joystick/joystick_driver.go
```diff
@@ -31,6 +31,12 @@ const (
 
 	// Configuration for the Xbox One controller.
 	XboxOne = "xboxOne"
+
+	// Nvidia Shield TV Controller
+	Shield = "shield"
```
adregner, line 36: seems this was pulled in when I was rebasing release -> dev because I accidentally started there.

platforms/joystick/joystick_nintendo_joycon.go
```diff
@@ -0,0 +1,115 @@
...
+			ID:   19,
+		},
+		pair{
+			Name: "sr_right",
+			ID:   16,
+		},
+		pair{
+			Name: "sl_right",
+			ID:   18,
+		},
```
adregner, line 104: these are the "inner" right/left buttons usually only used when used as a single controller, and not with two paired together.

gen2thomas:
Hi @adregner , thanks for your work and contributing to gobot. I will make a review in the next days. For now: the nvidia shield part was accidentally removed by me with this #871 . Very nice that this was detected and fixed by you. Please could you fix also the wrong name of "sheild.json" and mention this change in your PR description.
Thanks!

gen2thomas (commented):
It seems some files/changes are missing:
* please add your driver to the README.md of joystick folder
* there is no json file for this joystick (I just see it in the example, I have never tested any joystick and according to README.md the file is not mandatory)

adregner:
I think i've updated it as intended.  I also just found my "Nintendo Switch Pro Controller" and it seems to use the same codes as this, except there are at least 2 fewer buttons.  I'm still playing with it but this should be fine for the JoyCon's.

gen2thomas:
Thanks again. Now I will going to merge it.

Issue #920, opened 2023-05-19
installation problems with "v2"


gen2thomas:
latest recognized version seems to be 1.16.0: 
```
go install gobot.io/x/gobot@latest
go: downloading gobot.io/x/gobot v1.16.0

```
try to install version 2.0.0 will result in an error:
```
go install gobot.io/x/gobot@v2.0.0
go: gobot.io/x/gobot@v2.0.0: gobot.io/x/gobot@v2.0.0: invalid version: module contains a go.mod file, so module path must match major version ("gobot.io/x/gobot/v2"
```
For root cause, see e.g.: https://research.swtch.com/vgo-import

the solution: https://github.com/golang/go/wiki/Modules#semantic-import-versioning

example: https://github.com/cpuguy83/go-md2man/blob/master/go.mod

gen2thomas:
Version v2.0.1 is released to fix this problem, but it don't. The new version can not be found with an external example project.

```
go list -u -m all
go: unrecognized import path "gobot.io/x/gobot/v2": reading https://gobot.io/x/gobot/v2?go-get=1: 404 Not Found
go: gobot.io/x/gobot/v2@v2.0.1: missing go.sum entry; to add it:
	go mod download gobot.io/x/gobot/v2
```

I have found different resources for v2+ modules:
* https://go.dev/blog/v2-go-modules --> suggests to create a folder "v2" with a copy of code, due to best backward compatibility
* https://donatstudios.com/Go-v2-Modules --> outlines the history of v2+
* https://therootcompany.com/blog/bump-go-package-to-v2/ --> the result can be a caching problem
* working example - there are 52 versions now: https://github.com/google/go-github

gen2thomas:
v1:
```
go list -m -json gobot.io/x/gobot@latest
{
	"Path": "gobot.io/x/gobot",
	"Version": "v1.16.0",
	"Query": "latest",
	"Time": "2022-05-02T19:56:19Z",
	"Dir": "/home/user/go/pkg/mod/gobot.io/x/gobot@v1.16.0",
	"GoMod": "/home/user/go/pkg/mod/cache/download/gobot.io/x/gobot/@v/v1.16.0.mod",
	"GoVersion": "1.13"
}
```
v2+:
```
go list -m -json gobot.io/x/gobot/v2@latest
[...]

PR #947, opened 2023-06-04
all(bluetooth): adaptors removed to prevent usage of tinygo

gen2thomas:

This is just a last resort to make module mode working with v2+ release, see #920 . Affected:
* BLE
* BLE-firmata
* Microbit
* Parrot-Minidrone
* Sphero-BB8, Ollie, Sprkplus

## Manual test

- Linux

## Checklist

- [x] The PR's target branch is 'hybridgroup:dev'
- [x] New and existing unit tests pass locally with my changes (e.g. by run `make test`)
- [x] No linter errors exist locally (e.g. by run `make fmt_check`)
- [x] I have performed a self-review of my own code

If this is a removed driver or adaptor:

- [x] I have removed the names from the README.md
- [x] I have removed examples
- [x] I have run `make examples_check`


deadprogram:
This is part of Gobot I really quite need. :crying_cat_face:

gen2thomas:
Hi @deadprogram , by prevention of "tinygo/bluetooth" we will lose support for some adaptors, more than I thought. I'm also sad about this. Nevertheless I would test the effect for release of v2+. 

If it works, I will create a new issue for restoring the files. Maybe we can implement a mock package in this case, so the user is able to fast switch to the real implementation. 

If the test is not working, I will restore it immediately and go forward with the v3 approach.

I would do this (merge , release, test, restore/issue) within one or two days until end of this week.

deadprogram:
I guess I still don't fully understand the issue. I thought I added the bluetooth package to a previous release before the whole 2.0 thing?

deadprogram:
See https://github.com/hybridgroup/gobot/issues/920#issuecomment-1578459073

I think this PR here is no longer needed.

gen2thomas:
You are right, unfortunately we have not detected/fixed the root cause yet. Lets see what happen with v3 - if the problem still exist, I will try to implement a mock for tinygo/bluetooth.

Issue #952, opened 2023-06-12
dependency "github.com/pkg/errors" can be avoided


gen2thomas:
It is used by accident in "adxl345_driver.go" and "mqtt_adaptor.go". The usage of "errors.Wrap()" in "ble_client_adaptor.go" can be substituted by standard Go.

PR #958, opened 2023-06-14
ble: simplify and substitute errors.Wrap()

gen2thomas:

Archived package ""github.com/pkg/errors" now removed from gobot, see #952 . Additionally:
* multiple wrapping of the same error was removed in "ble_client_adaptor.go".
* go.mod and go.sum cleaned

## Manual test

no

## Checklist

- [x] The PR's target branch is 'hybridgroup:dev'
- [x] New and existing unit tests pass locally with my changes (e.g. by run `make test`)
- [x] No linter errors exist locally (e.g. by run `make fmt_check`)
- [x] I have performed a self-review of my own code


platforms/mqtt/mqtt_adaptor.go
```diff
@@ -3,17 +3,17 @@ package mqtt
 import (
 	"crypto/tls"
 	"crypto/x509"
+	"fmt"
 	"os"
 
 	"gobot.io/x/gobot/v2"
 
 	paho "github.com/eclipse/paho.mqtt.golang"
-	"github.com/pkg/errors"
```
jney, line 11: Can't it be just "errors"?

gen2thomas: yes, this would be also possible, but a good practice is to use one of "fmt.Errorf()" or "errors.New()" in one file (some say also in one project)

